PR #65, opened 2019-04-13
tbl_regression - print interactions

ddsjoberg:
This is a major refactoring of the `tbl_regression` function.  The code has been updated to accurately parse interaction terms from model results.

@karissawhiting @michaelcurry1123 if you could review before I merge with the dev branch that would be great.  The branch passes the RMD Checks and the unit tests, but a second pair of eyes will be invaluable for this major change 🥇


michaelcurry1123:
@ddsjoberg  this is great! I was not able to break it but I did have one question. Some interaction terms I was producing were giving NA because the combination did not exist in the data and they were not the reference group. Did you specifically not want the table to show NA values when the estimate did not exist.

Here is an example  👍 
```
mt2 <- mutate(mtcars, 
              carbx = as.character(carb),
              cylx = as.character(cyl),
              gearx = as.character(gear))

tbl_regression(lm(mpg~(cylx +carbx)^2,mt2))
```
image

ddsjoberg:
Thanks @michaelcurry1123 !  I had no idea you could include interaction terms in this way haha

I did not mean to omit them.  It looks like the omission occurs in the `broom::tidy()` step:
image

So, I think we'll just have to leave it as is (which i think is fine)!

ddsjoberg:
If you think it's ready can you "Approve" the PR, and I'll merge it in?

ddsjoberg:
I got excited, and bypassed the official review and merged it

michaelcurry1123:
No problem, also fun fact if you add a third variable like this `tbl_regression(lm(mpg~(cylx +carbx+vs)^2,mt2))` it will give all 2 way interactions and you can change the ^2 to ^3 to get the 3 way interaction, all two way interactions and main effects.

Issue #27, opened 2019-03-09
include figures in {gtsummary} help files


ddsjoberg:
The gt output tables don't display in the online help files.

Look at the gt package help pages to see who they display their examples.  It looks like they are displaying image files.
Perhaps one solution is to add all the code to create the tables in the data-raw folder, and use code to save image of the tables.  The help files will then import the images.

ddsjoberg:
Updates to the {gt} package that will hopefully will be live soon, will help make the documentation of {gtsummary} results much easier.
https://github.com/rstudio/gt/pull/257

ddsjoberg:
Is there a way to incorporate the examples in the README.rmd file in this automated process?

karissawhiting:
Yes, that should be doable.

PR #83, opened 2019-04-22
Adds figures to documentation and script to automatically update them as needed

karissawhiting:
Adds figures to docs and script to generate them. Does not update README figures yet. Figure sizes may need to be adjusted. Re: Issue #27


ddsjoberg:
OMG this PR is so so amazing!  I can't wait to review :)

ddsjoberg:
The tables look amazing!  The code you wrote to create them is so short and easy to follow!

The checks all run without error/warning, and the help files all contain the amazing tables.  A few initial thoughts (all of these can be addressed in a separate PR):
1. The tables look especially nice on the {pkgdown} site :)
1. I think we should shorten the `tbl_summary` table examples (and perhaps others).  Also, two examples is should be sufficient (rather than the three we have).  With fewer examples and shorter tables, the space between example code and the rest of the help file will be shorter.  This requires no change before merging the pull request.  We can make a separate issue to make these kinds of updates. 
1. Rather than call the section "Figures", perhaps we should call it "Example Output"?  That may help people understand more easily that these are the output tables.
1. Is there a way to make a separation between two example tables?  Maybe an html break, \<br\>? 
1. This last one is a me issue I've had with {webshot}.  When I try to run the code, I get an error.  Do you know how to resolve it?
 
image

Issue #97, opened 2019-04-27
unify naming conventions


ddsjoberg:
There is unfortunately no unifying naming convention in the package.  For example, in `tbl_regression` the upper and lower limits of the confidence interval are called `ll` and `ul`.  In `tbl_survival` they are named `lower` and `upper`.

I am thinking the names should all match the output of `broom::tidy` for internal and external consistency.
```
broom::tidy(
  lm(hp ~ mpg, mtcars),
  conf.int = TRUE
)
#> # A tibble: 2 x 7
#>   term        estimate std.error statistic  p.value conf.low conf.high
#>   <chr>          <dbl>     <dbl>     <dbl>    <dbl>    <dbl>     <dbl>
#> 1 (Intercept)   324.       27.4      11.8  8.25e-13    268.     380.  
#> 2 mpg            -8.83      1.31     -6.74 1.79e- 7    -11.5     -6.16
```

This would also mean I would need to change the `coef` column name to `estimate`, and all p-value columns to `p.value` (I think that are all `pvalue` currently).

ddsjoberg:
Updates to make
- All p-values should be called `p.value`
  - [ ] `add_comparison`
  - [ ] `tbl_regression`
  - [ ] `tbl_uvregression`
  - [ ] update references to p-values in `inline_text`
  - [ ] `tbl_merge`
  - [ ] `add_global`
  - [ ] `add_q`
  - [ ] `tab_style_bold_p`
- Lower and upper bounds of confidence intervals should be called `conf.low` and `conf.high`
  - [ ] `tbl_regression`
  - [ ] `tbl_uvregression`
  - [ ] `tbl_survival`
  - [ ] update references to limits in `inline_text`
  - [ ] `tbl_merge`
- Primary estimates should be called `estimate` (`coef` -> `estimate`)
  - [ ] `tbl_survival`
  - [ ] `tbl_regression`
  - [ ] `tbl_uvregression`
  - [ ] update references to estimates in `inline_text`
  - [ ] `tbl_merge`

PR #100, opened 2019-04-27
#97 naming conventions

ddsjoberg:
#97


ddsjoberg:
Hey @karissawhiting and @michaelcurry1123 ,  

I wanted to flag this update as it may break some existing code.  I updated the column names in the `.$table_body` output objects to be the same for all functions.  All p-values are now `p.value`, lower confidence interval bound is `conf.low`, upper bounds are `conf.high`, coefficients from regression models are now `estimate`.  These names match the output of the `broom::tidy()` output.  Going forward, I think it will be best to have a unified naming convention.  Hope it doesn't cause any headaches for you!

Issue #146, opened 2019-05-15
Add support for ..numeric.., ..integer.., etc for specifying column attributes in tbl_summary


sammo3182:
A quick question-suggestions:

Is there a way to change the categories of variables without specifying each of them? In `fmt_table1`, variables will be presented according to their properties (continuous or categorical/binary). You offered a way to customize that with the argument type in the current version of the function. What I was looking for was a way to change how `fmt_table1` treats a whole class of variables. For instance, based on my limited practice of this function, it appears to treat integer vectors with a small number of values as "categorical." So, is there a way that I can ask the function to treat all types of numeric vectors as "continuous," no matter how many types of values it has---a way that I don't have to specify each vector through the type argument?

ddsjoberg:
Hi Sammo, 

I like your suggestion a lot.  Imagine I have a data set with columns that are integers and doubles, and we want all integer columns to be displayed as categorical.  Perhaps something like this could work, where `..integer..`, `..double..`, `..numeric..`, etc. are special names.
```r
my_data %>%
  fmt_table1(type = list(..integer.. = "categorical"))
```
If I wanted all numeric variables to be displayed as categorical variables in the table, I would use:
```r
my_data %>%
  fmt_table1(type = list(..numeric.. = "categorical"))
```

Does this look like what you have in mind?  

I think you're using v0.1.0.  I have implemented a lot of updates in v1.0.0 (not yet on CRAN).  In this version, I reserve the names `..continuous..`, `..categorical..`, and `..dichotomous..` to make changes to the class of variables in one statement.  But I hadn't thought of your suggestion, and I like it.

sammo3182:
It will be great if the function has that argument as you described! An alternative approach might be a separate function mimicking `dplyr::mutate_if` to define the presenting types before running `fmt_table1`. Something like: 

```r
my_data %>% 
  fmt_categorize_if(is.integer, "categorical") %>% 
  fmt_table1()   
```

An advantage of this approach is you don't need to reserve names for every class of vectors. Users are also more free to customize the presenting rules, such as only when integer variables have fewer than three unique values is the "categorical" rule applied (` fmt_categorize_if(length(unique(.)) <= 3, "categorical") `). 

Of course, I just propose this suggestion from the perspective of users. It might be super hard to achieve from the developer view. So, any effort to address this suggestion will be sincerely appreciated!

ddsjoberg:
I'll add to my list!

[...]

PR #162, opened 2019-06-14
146 tidyselect

ddsjoberg:
The selector functions are ready!  `all_numeric()`, `all_double()`, `all_integer()`, `all_logical()`, `all_character()`, `all_continuous()`, `all_categorical(dichotomous = TRUE)`, and `all_dichotomous()`.  We can also use all of the {tidyselect} functions, as well as `vars()` from {dplyr}.

This is a pretty big change.  Can you do a pretty thorough review?  I think everything should be good to go.  One thing that should probably be added is a brief explanation of the select functions.  Perhaps a vignette showing how to use them, and link to the vignette in the help files of functions that use them?


ddsjoberg:
Issue #146 lists each of the updates made in the package

ddsjoberg:
Are there any other selectors that we need?

ddsjoberg:
@karissawhiting when you build the package on your mac, do you get these warning messages?

image

I can't figure out how to get rid of them, but after some searching it may be a Windows thing

karissawhiting:
@ddsjoberg I am not getting those warning messages building on mac.

ddsjoberg:
@karissawhiting ok thanks!  i wish i weren't seeing them, but i guess it's ok :)

karissawhiting:
This functionality is great! 

A note about the new formula syntax. When I use the old syntax: 

```r 
trial %>%
    dplyr::select(age, response, grade) %>%
    tbl_summary(statistic = list(age = "Median {median}"))
```
I am not getting a deprecation message. Also, when I use the old and new syntax together (probably rare that this will happen), the new formula syntax part is ignored.

Example:

```r
iris %>%
  tbl_summary(statistic = list(all_continuous() ~ "{median}", 
                               "Sepal.Length" = "{mean}"))
```

ignores `all_continuous() ~ "{median}"`, but respects ` "Sepal.Length" = "{mean}"`.

ddsjoberg:
Thank you for the thorough review!!!!

I looked into it, and I was using the incorrect deprecation function (one that only displays the message when the code is called from the global environment).  Updated the code to display the message accurately.

For the second point when the old a new method is mixed...the code checks is the passed list has names (the old syntax).  If there are names it uses the old syntax.
image
The list passed in the example does have names and only uses the named pieces.  I could add another check that the names are all non-empty strings...but I am leaning toward not adding checks for the old syntax.  What do you think?

ddsjoberg:
OK!  Thank you so much for the review.  The deprecation message has been updated (along with some documentation updates, which led to so many documentation files being updated).

ddsjoberg:
@karissawhiting @margarethannum ready for another lookie if you can :)

ddsjoberg:
@margarethannum that's a great update!  It's now included in the PR branch :100:

margarethannum:
I have one last suggestion which is more philosophical, and may impose too much style on the user. 

BUT to highlight some of the usability of tidyselectors (in both {gt} and {gtsummary}, would you consider, as part of this update, renaming the pathology variables "stage" and "grade"  as "path_stage" and "path_grade" as part of the workflow I suggested above? Then we can highlight really using something like `starts_with("path")`. To my above selection we can add something cool from the {gt} package: 

```
trial %>%
  dplyr::select(trt, response, age,  path_stage = stage, marker, path_grade = grade) %>%
  tbl_summary(
    by = "trt",
    type = list(c("response", "grade") ~ "categorical"), # select by variables in a vector
    statistic = list(all_continuous() ~ "{mean} ({sd})", all_categorical() ~ "{p}%") # select by summary type/attribute
  ) %>%
  add_p(test = list(contains("response") ~ "fisher.test", # select using functions in tidyselect
                    all_continuous() ~ "t.test")) %>%
  as_gt() %>%
  tab_row_group(
     group = "Pathology",
rows = starts_with("path"))


```

- I think I need to fix above code a little but you get the idea?

ddsjoberg:
I like that a lot for illustrating the {tidyselect} functions.  The only reason i am hesitant is that the variables names are now colloquial and understood by everyone, and the update is slightly less understandable (and I am feeling some {gtsummary} fatigue!).  We could change the example, maybe?  Do you know of any other data sets in base R (or survival, or gt, or any other package that is loaded with gtsummary) with variables that have common prefix?

PR #195, opened 2019-08-07
Prepare for tidyr v1.0.0

jennybc:
I'm working on revdep checks as we work towards the release of tidyr v1.0.0. We see this for gtsummary:

https://github.com/tidyverse/tidyr/blob/master/revdep/problems.md#gtsummary

This PR will work with the current version of tidyr (v0.8.3), but also with the changes coming in what's slated to become v1.0.0. After the tidyr release dust has settled, you might eventually want to bump the minimum version and take this conditional code back out.

### DIFF
--- a/R/gtsummary-package.R
+++ b/R/gtsummary-package.R
@@ -10,3 +10,7 @@
 
 # allowing for the use of the dot when piping
 utils::globalVariables(c("."))
+
+tidyr_has_legacy_nest <- function() {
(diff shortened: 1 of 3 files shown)


jennybc:
BTW I should also let you know that this form of tidyr version check will generate a NOTE. This approach and how to explain in to CRAN is described more here:

https://principles.tidyverse.org/changes-multivers.html

ddsjoberg:
Thank you for the heads up and the updates!  Happy Coding :)

Issue #194, opened 2019-08-07
add_p() - let users specify custom functions for tests


ddsjoberg:
Please briefly describe your problem and what output you expect.

Please include a minimal reproducible example (AKA a reprex). If you've never heard of a [reprex](https://reprex.tidyverse.org/) before, start by reading <https://www.tidyverse.org/help/#reprex>.

---

Brief description of the problem

```r
# insert reprex here
```

ddsjoberg:
PR Merged

PR #204, opened 2019-08-08
add_p() - let users specify custom functions for tests 

ddsjoberg:
Issue #194

Externally, there are no changes to the API.  Updates were made to the code that return p-values.  It's now setup as an S3 Generic Function so users can write their own functions for calculating p-values.

This functionality has yet to be documented.  Here's an example of defining your own function.
```r
library(gtsummary)

# argument that can be passed: 
# data, variable (string), by (string), group (string), 
# type (string that is equal to one of 'continuous', 'categorical', 'dichotomous')

# this bad test always returns 0.05
add_p_test.sigtest <- function(...){
  0.05
}

trial %>%
  tbl_summary(by = trt) %>%
  add_p(test = everything() ~ "mytest")

# this bad test returns a random number
add_p_test.random_test <- function(...){
  runif(1)
}

trial %>%
  tbl_summary(by = trt) %>%
  add_p(test = everything() ~ "random_test")
```


ddsjoberg:
@zabore I made an update to the p-value functions you wrote.  Would you mind taking a look before we merge into the dev branch?

Issue #331, opened 2020-01-02
deprecate `exclude=` arguments


ddsjoberg:
Many functions have both `include=` and `exclude=` arguments (e.g. `add_p()`, `tbl_regression()`, `add_global_p()`, `tbl_uvregression()`, `as_gt()`, `as_kable()`, `as_tibble()`, etc.).  But we now support robust tidyselect notation and combining include/exclude into a single argument is now feasible.

Should the exclude argument be deprecated?
@karissawhiting @margarethannum 

``` r

library(gtsummary)
# only include grade
tbl_uvregression(
  trial,
  method = lm,
  y = age,
  include = grade
) %>%
  purrr::pluck("table_body", "variable") %>%
  unique()
#> [1] "grade"

# include all variable, except grade
tbl_uvregression(
  trial,
  method = lm,
  y = age,
  include = -grade
) %>%
  purrr::pluck("table_body", "variable") %>%
  unique()
#> [1] "trt"      "marker"   "stage"    "response" "death"    "ttdeath"
```

<sup>Created on 2020-01-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

*We would not need to deprecate the argument in `add_global_p.tbl_regression()` because it was added after the last CRAN release.*

karissawhiting:
Agree, I think it is cleaner with just `include=`, and also more in line with tidyverse (e.g. `select()`)!

PR #333, opened 2020-01-05
deprecated exclude argument

ddsjoberg:
The `exclude=` argument, used throughout the package has been deprecated.  PR also includes various updates to function documentation.

#331

Checklist for PR reviewer

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,7 @@
-- What changes are proposed in this pull request?
+**What changes are proposed in this pull request?**
 
-- If there is an GitHub issue associated with this pull request, please provide link.
+
+**If there is an GitHub issue associated with this pull request, please provide link.**
 
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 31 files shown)


ddsjoberg:
@margarethannum i was thinking about these updates...the gtsummary selectors are available for every input by default (e.g. `all_interger()`, `all_factor()`, etc).  BUT, the args in `tbl_regression()` can't use these selectors, since the regression terms don't have dataframe -like classes.  I'll need to make an update to not use those select helpers in those cases.   

Also, in the next release of tidyselect, we will be able to use predicate selectors like, `is.numeric` which will be weird in these settings....hmmmm....maybe we can have a chat to discuss if ALLLL inputs really need these flexible inputs?  i am not sure!

margarethannum:
I'm working on some other documentation updates to try and make the homepage more concise. 

I realize we never show a user how to use the {labelled} package in any of the tutorials, do you think we should do this? Since our `trial` dataset already has labels it never comes up, although you mention in text that we recommend using the {labelled} package, wondering if it should come up in any of the vignettes or gallery?

Issue #257, opened 2019-10-25
New function tbl_crosstab() ?


ddsjoberg:
Should we create a new function that simply cross tabulates data and prints in a nicely formatted table?

There would be a lot of overlap with `tbl_summary()` but there are certainly times a need a more basic  cross tab table that is not possible with  `tbl_summary()` .  Something to think on....

michaelcurry1123:
I know I commented on this in slack but the janitor package has an `adorn_crosstab()` function that has this functionality.  https://cran.r-project.org/web/packages/janitor/janitor.pdf . I think it integrates well tidyverse.

ddsjoberg:
@michaelcurry1123 the janitor package's functions are pretty great, and the results are pretty close to what anyone would need. I didn't see easy ways to modify the statistics presented (rounding, etc.).

@karissawhiting you mentioned wanting this functionality recently.  Would janitor be good enough for what you need?

One possible solution that would be pretty easy to implement, would be to add a function `add_margin()` that would
- Check the input is indeed an n by m table.
- Run `add_overall()` to add the row margins
- Calculate the column margins and add those statistics as well.

We could include an option whether to add row AND column margins, or either one only.  We could right it in such a way that it would already be compatible with `inline_text()`.

michaelcurry1123:
@karissawhiting @ddsjoberg  if the janitor package isn't sufficient we might be able to modify the functions to add in additional functionality (e.g. modify stats or rounding and append on methods so that gtsummary accepts them?)  this is more of a question than anything else.

karissawhiting:
@ddsjoberg @michaelcurry1123 these janitor options are great! Sidenote: It looks like `adorn_crosstab()` is deprecated in favor of `tabyl()` plus the `adorn_*()` functions.  

I do still think there would be value in adding crosstable functionality to gtsummary. This would allow users to leverage some existing gtsummary features for their crosstables, like `inline_text()` and the use of labelled data. 

I like the solution Dan proposed. Additionally, I think we could consider adding a wrapper function with those options as default that also adds a `tab_spanner()`.

karissawhiting:
Discussed with @ddsjoberg , and decided to make a simple wrapper function `tbl_cross()` with arguments to chose margin totals to calculate (col, row, both), what types of percents to display (cell, row, col), and whether or not to include missing values. 

This will use `tbl_summary` and `add_overall()`with some light gt customization for default display.

ddsjoberg:
[...]

PR #369, opened 2020-02-01
257 add tbl_cross

karissawhiting:
This PR adds a simple `tbl_summary()` wrapper called`tbl_cross()` that returns a 2 variable cross table.

#257 

Checklist for PR reviewer

- [ ] Code coverage is suitable for any new functions/features. 
- [ ] Update `gt_sha` in `data-raw/gt_sha.R` AND run the file.


R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
+#' Create a cross table of summary statistics
+#'
+#' Wrapper for [tbl_summary] to cross tabulate two variables
```
ddsjoberg: Add experimental badge (can copy from `tbl_survival`)

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+#'
+#' @param data A data frame
```
ddsjoberg: Add `@inheritParams tbl_summary` to inherit parameter descriptions that are the same in the two functions

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+#' attribute label is `NULL`, the variable name will be used.
+#' @param statistic A statistic name in curly brackets to
+#' be replaced with the numeric statistic (see glue::glue).
+#' The default is `{n}`. If percent argument is `"column"`, `"row"`, or `"cell"`,
+#' default is `{n} ({p}%)`.
+#' @param percent Indicates the type of percentage to return if using the.
+#' `{p}` statistic. Must be one of
+#' `"column"`, `"row"`, or `"cell"`. Default is `"column"` when `{p}` is
+#' specified as `statistic`.
+#' @param missing Indicates whether to include counts of `NA` values in the table.
```
ddsjoberg: can you delete this and just inherit it from `tbl_summary()`?

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+#'
+tbl_cross <- function(data,
+                      row = NULL,
+                      col = NULL,
+                      label = NULL,
+                      statistic = NULL,
+                      percent = c("none", "column", "row", "cell"),
+                      missing = c("ifany", "always", "no"),
+                      missing_text = "Unknown") {
+  row_str <- var_input_to_string(
```
ddsjoberg: in all the other functions, we don't use the `*_str` suffix to denote the character version.  we just convert to character first thing...

what do you think of following the same naming convention?

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+    data = data,
+    select_input = !!rlang::enquo(row),
+    arg_name = "row_str",
```
ddsjoberg: this should be `arg_name = "row"`...this is used to create an informative error message for users when things go wrong.

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+    select_single = TRUE
+  )
+
+  col_str <- var_input_to_string(
+    data = data,
+    select_input = !!rlang::enquo(col),
+    arg_name = "col_str",
```
ddsjoberg: same as above

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+    select_single = TRUE
+  )
+
+  # matching arguments ---------------------------------------------------------
+  missing <- match.arg(missing)
+  percent <- match.arg(percent)
+
```
ddsjoberg: save the function inputs here and return them as a part of the return object (like the other functions)

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+
+  # if no x and y provided, default to first two col_strumns of data -----------
+  if (is.null(row_str) & is.null(col_str)) {
```
ddsjoberg: hmmmm, i am starting to think that perhaps we should just require the user to pass the variable names...how often do you really want the first two columns tabulated against one another?

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+    row_str <- names(data[, 1])
+    col_str <- names(data[, 2])
+  }
+
+  row_str <- row_str %||% names(select(data, -tidyselect::all_of(col_str)))[1]
```
ddsjoberg: if we do have defaults columns, i am leaning towards requiring both columns specified or neither.  it would just e more clear

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+  col_str <- col_str %||% names(select(data, -tidyselect::all_of(row_str)))[1]
+
+  # create new dummy col for tabulating column totals in cross table
+  data <- data %>%
+    dplyr::select(tidyselect::all_of(c(row_str, col_str))) %>%
+    mutate(Total = 1)
+
+  # get col label (for tab spanner)--------------------------------------------
+  label <- tidyselect_to_list(data, label) %||% NULL
```
ddsjoberg: you don't need the `%||% NULL` at the end.

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+
+  col_label <- label[[col_str]] %||%
+    attr(data[[col_str]], "label") %||% col_str
+
+  # statistic argument ---------------------------------------------------------
+
+  # if no user-defined stat, default to {n} if percent is "none"
+  statistic <- statistic %||% ifelse(percent == "none", "{n}", "{n} ({p}%)")
```
ddsjoberg: can we just delete this and pass `statistic = as.formula(glue("everything() ~ {statistic}"))` to `tbl_summary()`

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+  statistic_list <- rep(statistic, 3) %>%
+    as.list() %>%
+    set_names(names(data))
+
+  # omit missing data, or factorize missing level ------------------------------
+  lbls <- purrr::map(data, ~ attr(.x, "label"))
+
+  data <- data %>%
+    mutate_at(vars(row_str, col_str), ~ as.factor(.x)) %>%
```
ddsjoberg: there is no need to create the anon function here.

`mutate_at(vars(row_str, col_str), as.factor)`

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+        "always" = forcats::fct_explicit_na(., missing_text) %>%
+          forcats::fct_expand(missing_text)
+      )
+    )
+
+  if (missing == "no") {
+    data <- stats::na.omit(data)
+
+    message(glue(
+      "{sum(is.na(data))} observations missing data have been removed."
```
ddsjoberg: update this to `"{sum(is.na(data))} observations with missing data have been removed."`?

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+    attr(data[[i]], "label") <- lbls[[i]]
+  }
+
+
+  # create main table ----------------------------------------------------------
+  x <- data %>%
+    tbl_summary(
+      by = col_str,
+      statistic = statistic_list,
+      percent = switch((percent == "none") + 1, percent, NULL),
```
ddsjoberg: don't you wish there was a more elegant way of doing this?!?

what do you think of this instead?
`percent = switch(percent != "none", percent)`

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+      missing_text = missing_text
+    ) %>%
+    add_overall(last = TRUE) %>%
+    bold_labels() %>%
+    modify_header(
+      stat_by = "{level}",
+      stat_0 = " "
+    )
+
+  footnote_text <- footnote_stat_label(x$meta_data)
```
ddsjoberg: the correct footnote is already in the gt_calls list, right?  do we need this line?

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+  x$gt_calls[["tab_spanner"]] <-
+    glue(
+      "gt::tab_spanner(",
+      "label = gt::md('**{col_label}**'), ",
+      "columns = contains('stat_')) %>%",
+      "gt::tab_spanner(label = gt::md('**Total**'), ",
+      "columns = vars(stat_0))"
+    )
+
+  x$gt_calls[["tab_footnote"]] <-
```
ddsjoberg: i am not sure why this is needed?  isn't this returned from the `tbl_summary` call?

R/tbl_cross.R
```diff
@@ -0,0 +1,163 @@
...
+    glue(
+      "gt::tab_footnote(footnote = '{footnote_text}', ",
+      "locations = ",
+      "gt::cells_column_labels(columns = gt::vars(label)))"
+    )
+
```
ddsjoberg: also return a call object like the other functions (ensure the `tbl_summary` is not returned, since it was called as a helper not by the user)

R/tbl_cross.R
karissawhiting: Sure, that makes sense

R/tbl_cross.R
karissawhiting: Ah, I was wondering what that arg was doing :). Will update.

R/tbl_cross.R
karissawhiting: I'm still getting the hang of `switch()`, but I think that would work! Does that mean if percent does in fact equal "none, it will return NULL? When I try it alone it doesn't explicitly return `NULL` but I will test it in context!

R/tbl_cross.R
karissawhiting: This footnote business is so I can adjust the location of the footnote. Here I am overwriting the default tbl_summary() footnote to change the location and place it over "Characteristic" instead of the col labels because it applies to rows and columns both. There may be a more elegant way of doing this...

ddsjoberg: Hmmm maybe with `gt::tab_source_note()`? Then there is no need to place the footnote.

R/tbl_cross.R
karissawhiting: Agree that this is a bit messy, but I like the idea that you can do `trial %>% select(response, trt) %>% tbl_cross()` which mimics the behavior of `trial %>% select(response, trt) %>% table()`.

R/tbl_cross.R
karissawhiting: Yes, I was considering that. I'll give it a try in the next iteration.

karissawhiting:
@ddsjoberg  Do I need to add any kable calls?

ddsjoberg:
I don't think you need kable. They should all be returned by tblsummary, and the additional calls you're adding are not supported by kable

R/tbl_cross.R
ddsjoberg: That makes sense to me. Let's require to pass both names or neither?

ddsjoberg:
Merged in https://github.com/ddsjoberg/bstfun/pull/1

PR #433, opened 2020-04-04
added tbl_cross

ddsjoberg:
Migrated `tbl_cross` from bstfun to gtsummary.  There are a few differences from the bstfun version
- The p-value calculation was moved to a separate S3method, `add_p()`, e.g. `tbl_cross() %>% add_p()`
- Since the p-value is calculated in a new function, I changed the argument `pval_source_note=` to `source_note=`
- I added `inline_text.tbl_cross()` function.

FYI, I didn't add the functions to the pkgdown site, because there is another PR open that restructures the Reference page.  If that PR is merged first, then we'll need to add them in this PR.  If this PR is merged first, then we'll need to add the tbl_cross function in that PR (trying to avoid merge conflicts!)

-------------------------------------------------------------------------------

Checklist for PR reviewer


R/add_p.R
```diff
@@ -1,3 +1,14 @@
+#' Adds p-values to gtsummary table
```
karissawhiting, line 1: I love that this is a method function and works for tbl_cross!!

Is there any scenario we can think of that a user would want to use a custom p-value function and if so, should we support that?

R/tbl_cross.R
```diff
@@ -0,0 +1,169 @@
+#' Create a cross table of summary statistics
```
karissawhiting, line 1: Overall this looks great! It is much cleaner now that the p-value options are outside the function. 

One think I keep expecting to be able to do as a user is:

`trial %>% select(response, trt) %>%
  tbl_cross(col = trt) `

Do we think we should allow this?

R/inline_text.R
```diff
@@ -570,3 +570,53 @@ inline_text.tbl_survfit <-
...
+#' @examples
+#' tbl_cross <-
+#'   tbl_cross(trial, row = trt, col = response) %>%
+#'   add_p()
+#'
+#' inline_text(tbl_cross, row_level = "Drug A", col_level = "1")
+#' inline_text(tbl_cross, row_level = "Total", col_level = "1")
+#' inline_text(tbl_cross, col_level = "p.value")
+
+inline_text.tbl_cross <-
```
karissawhiting, line 597: It's great that this works for tbl_cross now 

I'm getting different behavior when I only select a `row_level` vs  when I only select a `col_level`. It might be nice if these were consistent (unless user selects `"p.value"`), or if there were a consistent/ more informative error. 

`tbl <- trial %>% tbl_cross(response, trt)`
`inline_text(tbl, col_level = "Drug A")`
`inline_text(tbl, row_level = 1)`

Maybe if the user only selects `row_level` but not `col_level`, we could return the total and vis versa for col?

The default is not listed on docs for col_level. 

Also, can we allow the user to specify `"Total"` for `col_level` too (along with `"stat_0"`)?

R/add_p.R
ddsjoberg: I think the API for custom p-values needs to change in the tbl_summary setting, so I am not keen to add the functionality here as well.  The way it's currently implemented, it's fragile and sometimes doesn't work.

R/tbl_cross.R
ddsjoberg: i think it makes the most sense to specify both or none.

R/inline_text.R
ddsjoberg: You're right this needs updating, but I am not sure what the best defaults should be?

1. being able to specify Total and stat_0 makes sense.
1. perhaps we should require both to row and col to be specified?  I don't want to default to total, since we'll add later the margin option to include/exclude the margins.

ddsjoberg:
Hmmm, I am thinking `inline_text()` needs more thought...there is custom code in there look for the text `"Total"`, but a user could easily change the label before the object is passed to `inline_text()`.

What do you think the the API should look like?  Should we leave `inline_text()` out of this release and work on it more later? @karissawhiting  I am not sure we have the time to dedicate to it this week?

PR #457, opened 2020-04-19
444 tbl cross margin

### LINKED ISSUE #444: Add `margin=` argument to `tbl_cross()`
Add an argument that allows user to request both margins, a single margin, or neither.

ddsjoberg:
- Added `margin=` arg to `tbl_cross()` allowing users to select row, column, both, or no margins...default is both.
- Deleted the `pattern=` arg from `inline_text.tbl_cross()`. There was an error since we convert both vars to factors, which led to errors when using the pattern arg.
- @karissawhiting I updated the second example in `tbl_cross()` to include percentages to show the functionality. If you don't like it, we can remove.

closes #444 

Checklist for PR reviewer

- [ ] Code coverage is suitable for any new functions/features. 


ddsjoberg:
> These changes look great! The `margin` argument is very nice, and agree `pattern` is not needed.
> 
> My only suggestion would be to maybe add an error message when the user specifies only a `col_level` or a `row_level` in `inline_text.tbl_cross()`. Currently, if only `col_level` is specified it returns an`NA`, but if only `row_level` is specified it returns `Error: Result must have length X, not 0 `. Another way we could handle this would be to return the margin totals if only one of the two arguments is specified.

When the user doesn't specify the row argument it defaults to the top row (where the p-value is), which is the same default in the tbl_summary version.  I think it's enough to just update the documentation saying the default it to return the top line?

karissawhiting:
Yea, it looks like when there is no `by` it does `NA`. When there is a `by` variable an error is thrown:

```
t2 <- tbl_summary(trial, by = trt) %>% 
add_p()

inline_text(t2, variable = grade)
#Error: Must specify `column` argument
```
It might be nice to have an error saying you have to specify both so the behavior is consistent between `col_level` and `row_level`?

This is minor and I'm fine the doc edit instead if you prefer.

PR #459, opened 2020-04-20
setting gtsummary themes

### LINKED ISSUE #424: gtsummary themes
This is still a vague idea....but I like the idea of being able to save and set themes for gtsummary.  A theme could be something as simple as setting the p-value formatting functions, but could be something more complicated like how continuous variables are displayed in summary tables, or the labels that are displayed in tbl_summary results.

- Formatting function defaults
  - e.g. set default p-value rounding function
 
- how continuous variables are formatted
  - the default is what we have now, with all the statistics on a single row
  - but users could request something that looks like this (and we could also come up with a way for users to pass statistics on multiple rows?):
  image
- How labels are displayed
  - For a variable label, we display the label, but perhaps we could allow users to change this for all variables.  For example, it could show the label and the statistic `"{label}, {statistics}"` would could resolve to "Grade, n (%)".
  - Could perhaps be expanded to include units, `"{label}, {units}" = "PSA, ng/ml"`, or `"{label}, {stat}" = "PSA, median (IQR)"`

ddsjoberg:
Adding a function `set_gtsummary_theme()`.  The help file has examples of the various options available. The file also includes a sample theme, `theme_gtsummary_jama()` that follows the JAMA reporting guidelines. 

closes #424 

To Do....
- [ ] Are we satisfied with the API? (this is the most important item!)
- [ ] Add ability to change the way continuous variables are presented; for example, on two lines with the label on top, and an indented row showing the statistic presented.  This doesn't really need to be done now.
- [x] Implement theme options for all functions in package
  - [x] tbl_summary family
  - [x] tbl_regression family
  - [x] tbl_uvregression family
  - [x] tbl_cross family
  - [x] tbl_survfit family
- [ ] What other items would we like to be able to control in a theme???

Checklist for PR reviewer


michaelcurry1123:
do you think it is worth the effort to rename `gtsummary_reset_theme()` to `reset_gtsummary_theme()` so it matches the same style as `set_gtsummary_theme()`

ddsjoberg:
@michaelcurry1123 that's a great suggestion.  Just updated!

michaelcurry1123:
When modifying the functions inputs it is kind of hard to know what types of function to use/or are the best to use. For example in  `pkgwide-fn:pvalue_fun` your example you use style_pvalue but you can also use round there instead. You can also get silly with it (see sample below). How would the user know what functions to use? Specifically this might be difficult when trying to set a theme that requires a function input. 

```

mylab <- gtsummary_theme_jama()


mylab2 <- list(function(x) round(x,4))

mylab3 <- list(function(x) paste0(x,"AAA"))

names(mylab2) <- names(mylab)[1]
names(mylab3) <- names(mylab)[1]
gtsummary::set_gtsummary_theme(mylab3)


trial %>%
  dplyr::select(age, grade, trt) %>%
  tbl_summary(by = trt) %>% 
  add_p() 
```

ddsjoberg:
- themes vignette added
- vignette replaces global option vignette in pkgdown site (global option vignette still exists, just no top level link to it)
- theme function names updated from `gtsummary_theme_*()` to `theme_gtsummary_*()`

PR #531, opened 2020-06-10
Summarize continuous vars with {N_obs} only in `tbl_summary()`

### LINKED ISSUE #458: tbl_summary(digits=) update
I think there is a better way to for this argument to work.

At the moment, the argument only controls continuous variables. If the continuous variable is summarized with three statistics (e.g. median, p25, and p75), users pass an integer to indicate the number of decimal places to round.  They can also pass three different digits, to round the variables to three different levels, e.g. `digits = age ~ c(0, 1, 2)`.

I want to expand it to accept value for categorical variables as well, but I am not sure the best way.

One possible method is to allow users to pass a function in the digits argument.  For example, if the statistic presented is just the percentage (`statistic = all_categorical() ~ "{p}"`), and they wanted all percentages formatted with the `style_sigfig()` function, they could pass `digits = all_categorical() ~ style_sigfig`.

Users would still be able to pass individual digits for rounding, and we could treat the digits as shortcuts for the `sprintf()` function, e.g. `function(x) sprintf(x, fmt='%#.1f')` for rounding a number to one decimal place.

An annoyance could occur for categorical variables, when the user wants to change the just the rounding for the percentage and not the other statistics: under the current system, they'd need to specify every statistic.  One way around this is to accept a named list, e.g. `digits = all_categorical() ~ list(p ~ style_sigfig)`; this would only change the percent column.  This would also allow users to change `N` and `n` formatting as well.

### LINKED ISSUE #528: tbl_summary using {N_obs} with continuous variables
I am getting an error when using {N_obs} by itself on continuous variables. For example:

```r
trial %>% tbl_summary(type = list("ttdeath" ~ "continuous"), statistic = list(all_continuous() ~ "{N_obs}"))
```
Results in the following error: 
```
Error in eval(parse(text = text, keep.source = FALSE), envir) : 
  object 'N_obs' not found
```

ddsjoberg:

- Users can now specify how many decimal places to round statistics for categorical variables, e.g. the Ns and the percentages. Use the `digits=` argument to specify the number of decimal places to round the stats using the same syntax for rounding the continuous variables statistics.

- This update fixes a bug allowing users to summarize continuous variables with {N_obs}, {p_miss}, etc only (i.e. no other summary statistics like mean or median).

closes #528
closes #458 

Checklist for PR reviewer


ddsjoberg:
When trying to resolve merge conflicts, I accidentally pushed the update directly to master that addresses #528 in commit 1c5a57574d8b73f6619d093eb68eaeb7a96d8cd8 . 

In the process I lost the changes associated with #458 (well, they are still there, but I need to spend time extracting them).  These changes overlap somewhat with another pull request #566 . I will add the updates addressing #458 after pull request #566 is merged.

Issue #536, opened 2020-06-17
add an option to customize "Overall" header when using add_overall()


larmarange:
When using `add_overall()` in a non-English table, it would be nice to have a very easy way to change the text of the header, something like `tbl %>% add_overall(header = "Total")`

ddsjoberg:
Hi @larmarange , I like your suggestion!  In the meantime, you can use:
```r
trial %>%
  tbl_summary(by = trt) %>%
  add_overall() %>%
  modify_header(stat_0 = "**Total**")
```

larmarange:
Hi @ddsjoberg   thanks for the feedback

Yes, I noted the possibility of using `modify_header()`. This last function allows deep customization, but not very intuitive when change changing the label. For example, in the example you provided, we lost the total number of observations.

ddsjoberg:
1. You can maintain the the N with 
```r
trial %>%
  tbl_summary(by = trt) %>%
  add_overall() %>%
  modify_header(stat_0 = "**Total** (N =  {N})")
```

2. We'll soon be releasing a language theme, where all the default printing can be in the language of your choice.  At the moment, it only supports Spanish, but it'll be pretty easy to add French!

larmarange:
Let me know. I would be happy to contribute to add French

ddsjoberg:
Thank you! The Spanish is under review at the moment, and when it's done we'll be ready to add other languages.

ddsjoberg:
Hello @larmarange !

We added a French theme, and it's in dev version of gtsummary now.  Would you mind checking it out?
http://www.danieldsjoberg.com/gtsummary/dev/reference/theme_gtsummary.html
http://www.danieldsjoberg.com/gtsummary/dev/articles/themes.html

larmarange:
I will look at it with pleasure.

Just before, I need to finalize a first draft of compatibility of  tbl_summary() with survey objects

ddsjoberg:
Awesome! That'll be great!
Here is each translation in a CSV file if that is easier to peruse https://github.com/ddsjoberg/gtsummary/blob/master/data-raw/gtsummary_translated.csv

ddsjoberg:
added 7/1/2020

PR #549, opened 2020-06-25
some corrections for French translation

larmarange:
Following request in #536 

Just a quick remark: to properly present a table in French, you should use comma as a decimal mark (and a space as a thousand separator).

Any plan to include the definition of decimal mark and thousand separator in `theme_gtsummary_language()`?

Regards


larmarange:
Other comment regarding French:

some punctuation like : or ; should have a non-breaking space (&nbsp;) juste before.

Should the punctuation be therefore part of the translation file?

larmarange:
Regrading the default theme, a comma is used to present IQR. As in some languages a comma could be used as a decimal operator, would it be better, for default, to use a dash (-) instead?

PR #554, opened 2020-06-26
Adding `type=` argument to `add_global_p()`

### LINKED ISSUE #553: Should add_global_p() use type II by default?
Currently, `add_global_p()` use type-III approach for calculating global p-values.

When there is nos interaction, type 2 and 3 are almost equivalent. 

If there are interactions, in general, if there is no significant interaction effect, then type II is more powerful, and follows the principle of marginality. If an interaction is present, then type II is inappropriate while type III can still be used, but results need to be interpreted with caution. BUT **it is necessary to set the contrasts option in R**. Because the multi-way ANOVA model is over-parameterised, it is necessary to choose a contrast setting that sums to zero, otherwise, the ANOVA analysis will give incorrect results with respect to the expected hypothesis. (**The default contrasts type does not satisfy this requirement**.)

cf. https://mcfromnz.wordpress.com/2011/03/02/anova-type-iiiiii-ss-explained/

Considering that less advanced users will use default contrasts and usually models without interaction, should it be better to use type II as a default value and let advance users switch to type III?

To be noted, it is currently not possible to change the type of p-value in `add_global_p()`

ddsjoberg:
- [x] Add `type=` argument to `add_global_p()` functions.
- [x] Add a theme element to change the default
- [x] Added a more detailed message listing the variables the global p-value is being calculated for. 
- [x] Added `include=` and `keep=` arguments to `add_global_p.tbl_uvregression()` to be consistent with `add_global_p.tbl_regression()`. The remaining inconsistency is that  `add_global_p.tbl_uvregression()` defaults to adding the p-value for all variables while  `add_global_p.tbl_regression()` defaults to categorical/interaction terms.  Hopefully the note about which variables are included is informative for users to see what is happening.

closes #553 

cc @larmarange

Checklist for PR reviewer


ddsjoberg:
Thanks @leejasme !  I added more test, and updated the message when no terms are selected.  What do you think?

ddsjoberg:
migrated these changes to https://github.com/ddsjoberg/gtsummary/pull/565

PR #572, opened 2020-07-10
Adding `add_p.tbl_survfit()`

### LINKED ISSUE #421: tbl_survfit support functions
- [x] `inline_text.tbl_survfit()`
- [x] `add_p()`
- [x] `add_n()`
- [x] `add_nevent()`

ddsjoberg:
- Major re-structuring of the `tbl_survfit()` internals.  This is in preparation for allowing users to create tables with multiple stratifying variables (e.g. investigate univariate analyses by sex, then by treatment, etc.).  The updated structure also makes it easier to add subsequent functions like `add_p()`.
- Transformed `tbl_survfit()` into a S3 function, with methods for list (a list of survfit objects), data.frame, and survfit.
    - The list method is the primary method. The other two merely set up the inputs into a list of survfit objects, and pass the result to `tbl_survfit.list()`
- Added `add_n.tbl_survfit()` and `add_nevent.tbl_survfit()` methods 
- Added the `add_p.tbl_survfit()` method! 
    - At the moment we do not yet support custom functions.
    - Only one type of test can be passed (thinking ahead to when there are multiple univariate analyses in one table). But I can't imagine when someone would want to mix the test type in a table full of univariate survival statistics.
    - I think more internal checks should be added. But first want a review of the general outline and API. Once we settle on a final API, we can include more consistency checks.

Other related items that still need to be done:
https://github.com/ddsjoberg/gtsummary/issues/589

closes #421

Checklist for PR reviewer


michaelcurry1123:
@ddsjoberg  should I update this branch before working on it? it is out of date with master

ddsjoberg:
@michaelcurry1123 good idea. I just hit the update button

ddsjoberg:
@michaelcurry1123 FYI, I updated `tbl_survfit()` to be more flexible with the inputs (details above), and did a little cleaning up of `add_p.tbl_survfit()`

michaelcurry1123:
For the logrank method rho does not do anything, which I think is correct. However you can still specify rho but it just seems that the argument is ignored. For the example below I specified rho as 80 to test this and the print out reads as follows: 

> 
> Calculating p-value with
> `survival::survdiff(formula = Surv(ttdeath, death) ~ grade, data = trial, rho = 0)`

My question is; should rho be there and if so should it always be 0 when specifying logrank? If it is always 0 should there be a message saying that the rho argument was ignored if it is non-zero when using the log rank test? this is not a major point but wanted to bring it up.  

code:
```
mod2 <- survfit(Surv(ttdeath,death)~grade , data = trial)

tbl_survfit(mod2, times = c(12,24)) %>% 
      add_p(test = "logrank", rho = 80)

```

ddsjoberg:
@michaelcurry1123 that is a good point. I think some messaging when the additional args are being ignored would be helpful. In this case, `rho = 0` is synonymous with the log-rank test which is why we're ignoring them. 

I think we need to decide on the API first, and then we can start making these improvements. What do you think?

michaelcurry1123:
no problem so to discuss the api questions above (my responses are in bold):
Do we ever need more than one test for a survfit table?  **I could see a scenario where if the proportional hazard assumption is violated someone would want to show both a log rank test and a p-value test that would address the issue of non-proportional hazards. What do you think? **
Should we update the API to match just for consistency across the package? **See below**
If we update to be consistent, then we'll need to remove add_p(footnote_text=) arg - **When using the app_p function it feels very similar to how it is implemented with tbl_summary. If the update is only to remove the footnote_text argument then I can see this going either way.   ** Let me know what you think @ddsjoberg !

ddsjoberg:
@michaelcurry1123 thank you for your thoughts!

I think you're right...consistent notation will be best throughout the package. 

Now for implementation...here is my initial thought.  Can it be improved upon?
- I like that the current method is flexible and allows users to pass additional args to `survdiff()`. To keep this functionality we can add a `test.args=` argument.
- The implementation could look something like this:
    -  `... %>% add_p(test = list(everything() ~ "survdiff"), test.args = list(everything ~ "list(rho = 0.5))")` (other conditions could be added to each of these lists)
    - We could also allow for simplified inputs like  `... %>% add_p(test = "survdiff", test.args = list(rho = 0.5))`

@karissawhiting do you have a preference for the API?

PR #629, opened 2020-08-20
#624 fix for NULL margin arg

### LINKED ISSUE #624: Bug in `tbl_cross(margin=)`
Unable to produce table with no margins.

```r
trial %>%
  tbl_cross(trt, response,
            label = response ~ "**Tumor Responded to Treatment**",
            percent = "row",
            margin = NULL)
```

image
FYI @karissawhiting

ddsjoberg:

- In `tbl_cross()` when the margin was left NULL it was incorrectly being defaulted to `"column"` in the `match.args()` fn

- Also removing the en dash as the default value for the missing value (responding to email from CRAN with encoding problem on Debian...effing debian)

closes #624 

Checklist for PR reviewer

- [x] Code coverage is suitable for any new functions/features. Review coverage with `covr::report()`. Before you run, set `Sys.setenv(NOT_CRAN=true)` and begin in a fresh R session without any packages loaded. 
- [x] NEWS.md has been updated with the changes from this pull request under the heading "`# gtsummary (development version)`". If there is an issue associated with the pull request, reference it in parantheses at the end update (see NEWS.md for examples).


ddsjoberg:
> In the `tbl_survfit()` docs the default argument is indicated as NULL but in the `margin` parameter text it says it is `"--"`. I can't think of a good way around this. I think it is okay as is.

this happens a few places throughout the package, so that we can allow the users to change the defaults using the themes.

PR #631, opened 2020-08-26
Adding continuous2 summary type

### LINKED ISSUE #620: Allow for multi-line summaries for continuous variables
- the summary type could be `"continuous2"`
- we could look for `"\n"` in the statistic to make the type `"continuous2"`, or maybe the statistic is a vector, e.g. `c("{mean}", "{min} -- {max}")`
- i think if there is at least one `"continuous2"` variable, we should default all continuous variables to this format.
- the output would look more like this:
    image

ddsjoberg:

This pull request adds another summary type to `tbl_summary()` called `"continuous2"`. This type adds a header label for continuous variables (similar to categorical types), and displays the statistics presented below.

```r
trial %>%
  select(age, marker, trt) %>%
  tbl_summary(
    by = trt,
    type = all_continuous() ~ "continuous2",
    statistic = list(all_continuous() ~ c("{median} ({p25}, {p75})", "{min}, {max}"))
  )
```

image

Question for @karissawhiting and @michaelcurry1123 : The statistic labels have been in lower case. I think that looked fine for the footnote, but doesn't look as good in these tables. What do you think? We could change the all the statistics (footnote and continuous2 labels).

PR to do list...
- [x] update `tbl_summary()` to accept `type = "continuous2"`
- [x] update `tbl_svysummary()` to accept `type = "continuous2"`
- [x] update `add_stat_label()` to handle continuous2 variables
    - allow users to use the `label=` argument to update continuous2 labels
    - users who select column location with continuous2 will have it changed to row
- [x] update `tbl_summary()` documentation
- [x] update `tbl_svysummary()` documentation
- [x] update `add_stat_label()` documentation to explain handling of continuous2 variables
- [x] improve error messaging when user passes vector of length >1 to statistic when type is not continuous2
- [x] add unit tests for all `tbl_summary()` family functions
- [x] add unit tests for all `tbl_svysummary()` family functions
- [x] add theme elements for default continuous2
- [x] should the stat labels now be capitalized by default?
- [x] add range to translations excel file
- [x] add `theme_gtsummary_continuous2()` to make continuous2 the default. Need to decide on a default statistic. Can probably add this as a an argument to the function.

closes #620

Checklist for PR reviewer

- [x] Code coverage is suitable for any new functions/features. Review coverage with `covr::report()`. Before you run, set `Sys.setenv(NOT_CRAN=true)` and begin in a fresh R session without any packages loaded. 
- [x] NEWS.md has been updated with the changes from this pull request under the heading "`# gtsummary (development version)`". If there is an issue associated with the pull request, reference it in parantheses at the end update (see NEWS.md for examples).


ddsjoberg:
Very first draft is ready.  Still more work to do...

```r
trial %>%
  select(age, grade) %>%
  tbl_summary(
    type = age ~ "continuous2",
    statistic = age ~ c("{median} ({p25}, {p75})", "{mean} ({sd})")
  )
```

image

ddsjoberg:
Hey hey @karissawhiting and @michaelcurry1123 ! This PR is ready for review. It resulted in quite a few updates throughout the package.  Before we merge could you both take a look so we don't introduce a bug 😬

michaelcurry1123:
@ddsjoberg  I will happy to take a look at this! I should be able to get to it tomorrow

michaelcurry1123:
@ddsjoberg this is great! I think I would default the capitalized, although it is easy to with `add_stat_labels()`. I checked the functionality with `tbl_merge()` and it works there as well but the labels do have to match exactly. Is that the intended effect you wanted?

ddsjoberg:
@michaelcurry1123 for `tbl_merge()`, yes, it merged on exact matches. Do you think something else would be preferable?

I'll update the stat labels to be upper case

michaelcurry1123:
@ddsjoberg  no I think that is fine and preferable. I can't think of a case where you wouldn't want to do that. This is the code I was thinking of but I don't think there is any realistic scenario that someone would want to do something like this:

```
tb1 <- trial %>%
  select(age, marker, trt) %>%
  tbl_summary(
    by = trt,
    type = all_continuous() ~ "continuous",
    statistic = list(all_continuous() ~ c("{median} ({p25}, {p75})"))
   , missing_text = "N/A"
  ) %>% 
  add_stat_label(label = all_continuous2() ~ c("Median (IQR)", "Range")) 


tb2<- trial %>%
  select(age, marker, trt) %>%
  tbl_summary(
    by = trt,
    type = all_continuous() ~ "continuous"
  ) 

tbl_merge(list(tb1, tb2))


```

ddsjoberg:
Merging after changing the label with add_stat_label() is an issue that has popped up before. But I think what we're doing in the most reasonable thing.....i hope!

ddsjoberg:
I'll update the `tbl_summary.Rmd` vignette with `"continuous2"` examples after this PR is merged.

karissawhiting:
This is a cool feature and the implementation looks really good! Agree with default capitalization of stat labels. 

I'm wondering if there is a way we can simplify this for users and avoid Introducing a new type (`continuous2`). If a user passes a vector to the RHS of the `statistic` formula that has length > 1, can we automatically default to multi-level? For example: 

```
# this would do multi-line stat
trial %>%
  select(age, marker, trt, response) %>%
  tbl_summary(
    by = trt,
    statistic = list(age ~ c("{median} ({p25}, {p75})", "{min}, {max}"))
  )

# this would do single line stat
trial %>%
  select(age, marker, trt, response) %>%
  tbl_summary(
    by = trt,
    statistic = list(age ~ c("{median} ({p25}, {p75}) ({min}, {max})"))
  )
```

I may be missing some important consequences of the above, so if we do proceed with having the `continuous2` type maybe we could call it something more intuitive (e.g.  `multi-continuous`, `multi-line-continous`, `labelled-continuous` or something). Also, what do we think about adding a default `continuous2` option that includes range so this:

```
trial %>%
  select(age, marker, trt, response) %>%
  tbl_summary(
    by = trt,
    type = all_continuous() ~ "continuous2")
```

would default show Median (IQR) \n Range for all continuous in table.

ddsjoberg:
@karissawhiting very good points. Let me think and I'll respond more fully.

The immediate thing that comes to mind if we choose to depend on the length of the statistics argument, is that we can't get multiple lines (label row and a stats row) if there is only one row of stats.

ddsjoberg:
@karissawhiting OK, gathering my thoughts a bit more now...

1. I think we need to have a new type; otherwise, there would be no way to show the results below
    image

1. Let's chat about the new type name. I don't think continuous2 is a great name. The others may be difficult to remember though?

1. Re: updating the default statistic..I think for the tables we have in mind the median IQR is sufficient.  I also don't like the range, because the max or min is often very skewed in our field and essentially meaningless when it comes to aggregate data statistics.  There are certainly there are instances where it is useful, but I am hesitant to make it a default.  Something I have thought about is writing a EDA theme (exploratory data analysis), where we would print verbose results with more results than we would typically (including min and max).

PR #725, opened 2020-12-24
Updates to `add_stat()`

### LINKED ISSUE #688: Genereralize `add_stat()` to work with `tbl_svysummary()` objects
https://stackoverflow.com/questions/64499552/using-gtsummary-to-display-confidence-intervals-for-survey-design-object

### LINKED ISSUE #714: Allow returning multiple values for multinomial variables with add_p and add_stat
**The idea**

I am looking for a way to use chi square or fisher tests on multinomial variables. Let's say I have a group variable (A or B) and an exposure variable (X, Y Z).

I'd like to compare Y and Z exposure with X between groups, which would generate a table like :

| Characteristic | Group A | Group B | OR | 95 CI | p.value |
|---|---|---|---|---|---|
|Exposure| | | |
|  X  |  75 | 23 | - | - | - |
|  Y  |  19 | 41 |1.5 | 1.1, 1.7 | 0.04 |
|  Z  |  103 | 81 | 2.7 | 2.5, 2.9 | 0.03 |

(random values here)

**Imagined solutions**
Similar tables are already possible with a tbl_uvregression() call (first level used as reference). At first I thought I should create a custom vetted model working with it. But it sounds wrong since it has nothing to do with a regression.

Other idea : allowing add_stat to display multiple p.values for a multinomial variable like this:

```r
my_fisher <- function(data, variable, by, ...) {
  # for example purpose only :
  list(
    X = list(or = NULL, ci = NULL, p=NULL), 
    Y = list(or = 1.5, ci = "1.1, 1.7", p=0.04), 
    Z = list(or = 2.7, ci = "2.5, 2.9", p=0.03) 
   )
}

my_data %>%
  select(group, exposure) %>%
  tbl_summary(by = group) %>%
  add_stat(
    fns = everything() ~ my_fisher,
)
```

For now it's not possible because add_stat is expecting my_fisher to return only one value. Would it be possible to allow returning multiple columns (or, ci and p here) and multiple rows (X, Y and Z here)?

ddsjoberg:
* The `add_stat()` function now supports `tbl_svysummary()` objects in addition to `tbl_summary()` (#688)
* Multi-dimensional statistics can now be added for categorical variables via `add_stat()`, i.e. one statistic for each level of the categorical level (#714)

closes #714
closes #688 

cc @JeremyPasco

Checklist for PR reviewer

- [ ] Approve Pull Request


JeremyPasco:
Hi, thanks for following up #714. Adding one statistic to each level of categorical variable with add_stat was definitly the main idea.
Just to be sure of the direction taken here: what if we want to add more than one statistic per level (ie: OR, CI and p-value)? Do you plan to allow this with a single call to add_stat, or should we call add_stat for each statistic ?
I'm asking this because since many statistics come "together" when produced in statistical test functions and it would make sense to me if we could bring all of them at once.

ddsjoberg:
> Hi, thanks for following up #714. Adding one statistic to each level of categorical variable with add_stat was definitly the main idea.
> Just to be sure of the direction taken here: what if we want to add more than one statistic per level (ie: OR, CI and p-value)? Do you plan to allow this with a single call to add_stat, or should we call add_stat for each statistic ?
> I'm asking this because since many statistics come "together" when produced in statistical test functions and it would make sense to me if we could bring all of them at once.

I certainly see where you're coming from, and I haven't thought about where to take it from here at all :)

Some initial thoughts:
1. Whatever we do should not be a breaking change to the current `add_stat()`.
2. It will become complex to pass n-dimensions of column headers, formatting functions, footnotes, etc.
3. Suggested Implementation: If the user were to pass a function that returns a data frame rather than a vector. We would need checks that the returned data frame matches the expected. This is not going to be an easy implementation for users I think, but can still be useful.....

ddsjoberg:
@JeremyPasco if you could please make a new issue for this and add suggestions an API that would be fantastic.

JeremyPasco:
Gladly! Here is my proposal #746

PR #748, opened 2021-01-14
Generalized add_stat() for more complex returns

ddsjoberg:
**What changes are proposed in this pull request?**
* The `add_stat()` function may now return multiple columns of new statistics. (#746)

**If there is an GitHub issue associated with this pull request, please provide link.**
closes #746 

```r
my_ttest <- function(data, variable, by, ...) {
  t.test(data[[variable]] ~ factor(data[[by]])) %>%
    broom::tidy() %>%
    select(estimate, conf.low, conf.high, p.value)
}

trial %>%
  select(trt, age, marker) %>%
  tbl_summary(by = trt, missing = "no") %>%
  add_stat(
    fns = everything() ~ my_ttest
  )
```
image

To do...
- [x] Update documentation
- [x] Test fn on categorical levels
- [x] Add more unit tests

cc @JeremyPasco FYI

--------------------------------------------------------------------------------

Checklist for PR reviewer

- [ ] PR branch has pulled the most recent updates from master branch. Ensure the pull request branch and your local version match and both have the latest updates from the master branch.
- [ ] If an update was made to `tbl_summary()`, was the same change implemented for `tbl_svysummary()`?
- [ ] If a new function was added, function included in `_pkgdown.yml`
- [ ] If a bug was fixed, a unit test was added for the bug check
- [ ] Run `pkgdown::build_site()`. Check the R console for errors, and review the rendered website.
- [ ] Code coverage is suitable for any new functions/features. Review coverage with `covr::report()`. Before you run, set `Sys.setenv(NOT_CRAN="true")` and begin in a fresh R session without any packages loaded. 
- [ ] R CMD Check runs without errors, warnings, and notes
- [ ] `usethis::use_spell_check()` runs with no spelling errors in documentation

When the branch is ready to be merged into master:
- [ ] Update NEWS.md with the changes from this pull request under the heading "`# gtsummary (development version)`". If there is an issue associated with the pull request, reference it in parantheses at the end update (see `NEWS.md` for examples).
- [ ] Increment the version number using `usethis::use_version(which = "dev")` 
- [ ] Run `codemetar::write_codemeta()`
- [ ] Run `usethis::use_spell_check()` again
- [ ] Approve Pull Request
- [ ] Merge the PR. Please use "Squash and merge".

PR #603, opened 2020-08-02
allowing for any class to be passed to tbl_summary

ddsjoberg:
**What changes are proposed in this pull request?**
This PR allows columns of any class to be summarized by `tbl_summary()`.

Outstanding items:
- [ ] Do more testing with dates, times, datetimes, and lubridates (including quarters within lunbridate)
- [ ] Can we improve error messaging if users sends column so a summary function that cannot handle it?
- [ ] Still need to remove custom ordered factors code. 
- [ ] You can pass dates now, but there is no way to format them 🤷 
- [ ] We open ourselves up to a whole new set of errors that are difficult to anticipate by accepting a column of any class. 
    - We need to add more informative error messaging, but that is actually hard to do.
    - We could add a column to `.$meta_data` indicating if the column is a base R type/class. When there are errors with non-base types, we could pass a special error message.

Haven labelled example
```r
tibble::tibble(
  s1 = haven::labelled(c("M", "M", "F"), c(Male = "M", Female = "F")), 
  s2 = haven::labelled(c(1, 1, 2), c(Male = 1, Female = 2)), 
  s3 = haven::labelled(c(1, 1, 2), c(Male = 1, Female = 2),
                       label="Assigned sex at birth")
) %>%
  tbl_summary(
    type = list(c(s2, s3) ~ "continuous")
  )
```

**If there is an GitHub issue associated with this pull request, please provide link.**

closes #569
closes #488
closes #540

--------------------------------------------------------------------------------

Checklist for PR reviewer

- [ ] PR branch has pulled the most recent updates from master branch. Ensure the pull request branch and your local version match and both have the latest updates from the master branch.
- [ ] If an update was made to `tbl_summary()`, was the same change implemented for `tbl_svysummary()`?
- [ ] If a new function was added, function included in `_pkgdown.yml`
- [ ] If a bug was fixed, a unit test was added for the bug check
- [ ] Run `pkgdown::build_site()`. Check the R console for errors, and review the rendered website.
- [ ] Code coverage is suitable for any new functions/features. Review coverage with `covr::report()`. Before you run, set `Sys.setenv(NOT_CRAN = TRUE)` and begin in a fresh R session without any packages loaded. 
- [ ] R CMD Check runs without errors, warnings, and notes
- [ ] NEWS.md has been updated with the changes from this pull request under the heading "`# gtsummary (development version)`". If there is an issue associated with the pull request, reference it in parantheses at the end update (see NEWS.md for examples).
[...]

PR #754, opened 2021-01-27
follow tidyverse style guide

IndrajeetPatil:
As outlined here: https://style.tidyverse.org/

Done using `styler` package: https://styler.r-lib.org/

### DIFF
--- a/R/add_difference.R
+++ b/R/add_difference.R
@@ -21,9 +21,11 @@
 #' add_difference_ex1 <-
 #'   trial %>%
 #'   select(trt, age, marker) %>%
-#'   tbl_summary(by = trt,
-#'               statistic = all_continuous() ~ "{mean} ({sd})",
-#'               missing = "no") %>%
+#'   tbl_summary(
+#'     by = trt,
+#'     statistic = all_continuous() ~ "{mean} ({sd})",
+#'     missing = "no"
+#'   ) %>%
 #'   add_n() %>%
 #'   add_difference()
 #'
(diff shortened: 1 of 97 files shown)


IndrajeetPatil:
I understand this is a massive change for a single PR, but I am not sure how else would you rather have me go about this. 

Making PR for every file doesn't make sense. So let me know what your preference is, and I can modify the PR accordingly.

ddsjoberg:
hey hey! so much style here! 😆 

the package was in need of doing this again...it had been quite some time since it was last run! it is on my list to implement automatic linters of the code to keep it nice, but haven't gotten around to it!

The only problem with merging this as is, is that there are two large PRs underway at the moment that make changes to many many many many files. I think if we merge now, it'll introduce merge conflicts 😨 

Once these two PRs are merged, we can do this one!
https://github.com/ddsjoberg/gtsummary/pull/748
https://github.com/ddsjoberg/gtsummary/pull/603

IndrajeetPatil:
Absolutely! I will close this for now and re-open later on.

I will submit another `PR` to change to `testthat` 3rd edition with parallel support to speed up testing. Is that okay?

ddsjoberg:
OMG I would love that! Thank you, thank you, thank you!

I think I use `context()` in every file, which I think is now deprecated, and probably a bunch of other stuff too that isn't good practice!  :stuck_out_tongue_winking_eye:

IndrajeetPatil:
Cool, will do that 😄

PR #756, opened 2021-01-27
adopts testthat 3 edition

IndrajeetPatil:
There are two tests in `test-add_p.tbl_svysummary` that fail, and I have no idea why! 

Otherwise, everything else is working as expected after these changes.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -111,3 +111,5 @@ Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1
+Config/testthat/edition: 3
+Config/testthat/parallel: true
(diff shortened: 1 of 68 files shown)


ddsjoberg:
WOW! Thank you so much! 🙌🏼 🙌🏼 🙌🏼

ddsjoberg:
I _think_ those new failures may be because waldo is using a relative difference rather than an absolute.   I can just increase the tol level...I'll look into it more this weekend!

ddsjoberg:
@IndrajeetPatil thank you again for this amazing PR!

There are a few more tests (all named `test-vetted_models-<model_name>.R`) that I have set to only run on GitHub Actions.  They still contain the deprecated `expect_equivalent()`.  I can update them if you don't want to.  I would just change them all to `expect_equal()`?

ddsjoberg:
@IndrajeetPatil 👐🏼 👐🏼 👐🏼 👐🏼 👐🏼 thank you so much for upgrading testthat!!

IndrajeetPatil:
Yaay!! 😊

PR #794, opened 2021-02-20
`tbl_summary()` will accept columns of any class

### LINKED ISSUE #488: Integration with `labelled` package `set_value_labels()` + `haven_labelled` class
Amazing package -- really love this project. I am trying to use it alongside with the `labelled` package and when using the `set_value_labels()` function I get an error:

```r
library(tidyverse)
library(gtsummary)
library(labelled)

mtcars %>%
  select(cyl, mpg) %>%
  set_variable_labels(cyl = "Cylinders",
                      mpg = "Miles per gallon") %>%
  set_value_labels(cyl = c("Four" = 4, "Six" = 6, "Eight" = 8)) %>%
  tbl_summary(by = cyl)
```

```
Column(s) ‘cyl’ omitted from output.
Accepted classes are ‘character’, ‘factor’, ‘numeric’, ‘logical’, ‘integer’, or ‘difftime’.
Error in class(data[[by]]) <- setdiff(class(data[[by]]), "labelled") : 
  attempt to set an attribute on NULL
```

Not sure if this intentional behavior for the package, or if it would be an easy fix. Using factor labels for the variable levels (like below) works, but it would be great if `gtsummary()` would also accept the `haven_labelled` class, as I'm seeing it used more and more. 

```
mtcars %>%
  select(cyl, mpg) %>%
  set_variable_labels(cyl = "Cylinders",
                      mpg = "Miles per gallon") %>%
  mutate(cyl = factor(cyl, labels = c("Four","Six","Eight"))) %>%
  tbl_summary(by = cyl)

```

### LINKED ISSUE #540: Update tbl_summary() internals to remove `tidyr::complete()`
`dplyr::group_by()` has a `.drop=` argument that can keep unobserved levels of a factor level.  There is much code dedicated to getting complete sets of results that can be much simpler to read! 

https://twitter.com/sharlagelfand/status/1275122060969287680

### LINKED ISSUE #569: `tbl_summary()` update for ordered factors and other class handling
`tbl_summary()` accepts ordered factors, but in a hacky way: we replace the ordered factor with a non-ordered factor.

```r
  # removing ordered class from factor variables -------------------------------
  data <- dplyr::mutate_if(data,
                           ~inherits(., "ordered") && inherits(., "factor"),
                           ~factor(., levels = attr(., "levels"), ordered = FALSE))
```

The downside to this approach is that the label is stripped from the column, and it's not clean code.

I would eventually like to update so that _all_ data types are accepted, including ordered factors.  The following steps would need to be taken.
- [ ]  Investigate the location int he code base that erred when ordered factors were passed, and update.
- [ ] Update to accept base R date and datetimes.
- [ ] The current reason we do not accept non-base R classes, is because we don't know how to summarize them.
  - For non-base R classes, should we print an error, and require the user to specify the summary type (continuous, categorical, dichotomous)?  Perhaps, for classes that we don't know how to summarize, we can omit them from the table and print a note that we didn't know how to summarize the non-base R class. Then direct them to specify the summary type OR convert to a base R class to have it included in the table.
  - [ ] Should we just guess and send the column through the defaulted summary functions? If we do this and there is an error, we should print an informative error message indicating the non-base R class passed, AND the function used to summarize. Ask the user if this class is supported in the summary function.

### LINKED ISSUE #780: value parameter in tbl_summary doesn't work on an empty category
### Expected behavior:

Setting the value parameter of tbl_summary on a category should display only this category, even if this category has a null count.

### Actual behavior

On categorical variables, setting the value parameter on a non represented category does nothing.

### Reprex

``` r
library(tidyverse)
library(gtsummary)

x <- trial %>% 
  transmute(
    trt = fct_expand(trt, 'Drug C')
  )

tbl_summary(x) # display all categories: OK
tbl_summary(x, value = trt ~ "Drug A") # display only Drug A: OK
tbl_summary(x, value = trt ~ "Drug C") # display all categories: wrong
```
<sup>Created on 2021-02-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

By this third summary, I expected one row with : `Drug C      0 (0%)`

ddsjoberg:

* `tbl_summary()` now accepts any class as input, previously only non-date base classes were accepted. For non-base R classes, the summary type must be specified using `tbl_summary(type=)`. The default summary statistic for dates/times is the minimum and maximum. (#488)

* Improved handling of ordered factors as the `by=` variable (#569, #540)

* Bug fix when displaying an unobserved level of a factor variable dichotomously in `tbl_summary()`. (#780)

closes #569 
closes #488
closes #540
closes #780

Checklist for PR reviewer

- [ ] Approve Pull Request


ddsjoberg:
@larmarange I am making an update that allows `tbl_summary()` to accept columns of any class. Is this something I should also incorporate with `tbl_svysummary()` or are the svy summary functions too specialized to handle other classes?

larmarange:
I have tested quickly. `svytable()` works with dates or times and seems pretty generic.

Therefore, `tbl_svysummary()` should follow `tbl_summary()`.

NB: internally, `svytable()` uses `xtabs()`

ddsjoberg:
Hey @larmarange ,

Neither the `quantile()` nor the `svyquantile()` functions work with dates. To get around this in `tbl_summary()`, I made the default statistics shown `statistic = "{min} to {max}"`. But this won't work with `tbl_svysummary()` since there are no `svymin()` and `svymax()` functions that work with dates.

There is a work-around where you `unclass()` the variable, calculate the quantile, then convert it back to its original class. Honestly, I don't work often with date/times in R, and when I do, I reference the lubridate cheat sheet. We could write a small wrapper for `quantile()` nor the `svyquantile()` to handle dates, but it does make me nervous I'll screw up a time zone or something that I am not aware of during the conversion back to the original class.

We can address this in a separate issue.

``` r
library(magrittr)

df_date <-
  data.frame(
    dates = as.Date("2021-02-20") + 1:10,
    times = as.POSIXct("2021-02-20 20:31:33 EST") + 1:10,
    group = 1:10 %% 2
  ) %>%
  {survey::svydesign(ids = ~1, data = ., weights = ~1)}

survey::svyquantile(
  design = df_date,
  na.rm = TRUE,
  keep.var = FALSE,
  quantiles = 0,
  x = ~dates
)
#> Warning in xy.coords(x, y, setLab = FALSE): NAs introduced by coercion
#> Error in h(simpleError(msg, call)): error in evaluating the argument 'x' in selecting a method for function 't': need at least two non-NA values to interpolate

survey::svyquantile(
  design = df_date,
  na.rm = TRUE,
  keep.var = FALSE,
  quantiles = 0,
  x = ~unclass(dates)
)
#>                    0
#> unclass(dates) 18679

survey::svyquantile(
  design = df_date,
  na.rm = TRUE,
  keep.var = FALSE,
  quantiles = 0,
  x = ~unclass(times)
) %>%
  as.POSIXct(origin = "1970-01-01 00:00:00 EST")
#> [1] "2021-02-20 20:31:34 EST"
```

<sup>Created on 2021-02-22 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

larmarange:
Minimum and Maximum does not depend on sampling weights or sampling design. They can therefore directly be calculated using `range()` by accessible the data variables (store in `$variables` in the design object).

``` r
library(survey)
#> Le chargement a nécessité le package : grid
#> Le chargement a nécessité le package : Matrix
#> Le chargement a nécessité le package : survival
#> 
#> Attachement du package : 'survey'
#> The following object is masked from 'package:graphics':
#> 
#>     dotchart
library(tidyverse)
df_date <-
  data.frame(
    dates = as.Date("2021-02-20") + 1:10,
    times = as.POSIXct("2021-02-20 20:31:33 EST") + 1:10,
    group = 1:10 %% 2
  ) %>%
  {survey::svydesign(ids = ~1, data = ., weights = ~1)}

svyrange <- function(v, na.rm = TRUE) {
  df_date %>% 
    purrr::pluck("variables") %>%   
    purrr::pluck(v) %>%
    range(na.rm = na.rm)
}
svyrange("dates")
#> [1] "2021-02-21" "2021-03-02"
svyrange("times")  
#> [1] "2021-02-20 20:31:34 CET" "2021-02-20 20:31:43 CET"
svyrange("group")
#> [1] 0 1
```

<sup>Created on 2021-02-23 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

ddsjoberg:
@larmarange FYI I updated and `tbl_svysummary()` does accept dates now. Just like `tbl_summary()` the default is to show the range, as the quantiles are not supported. There is one survey-related snag. I kept the current infrastructure within `tbl_svysummary()` that uses `svyby()` when there is a `tbl_svysummary(by=)` variable present. The `svyby()` strips variables of their date/time class and returns the numeric version (this does not happen without a by variable).  You can overcome the bad formatting by passing a specialized formatting function in `tbl_svysummary(digits=)`, and there is an example of this at the bottom of `test-tbl_svysummary()`.

larmarange:
Thanks @ddsjoberg

michaelcurry1123:
when you change the `statistic` argument for `all_continuous()`  you get an error for dates. Dates print the min and max with the default but in the below example it throws an error.  Perhaps dates should not be influenced by all_continuous? or should the user just put in the non-date variables to the statistic call below? 

```
library(gtsummary)

library(tidyverse)


trial <- mutate(trial, dates = seq(as.Date("1970-01-01"), as.Date("1970-01-01")+nrow(trial)-1,1),
                dates2 = sample(seq(as.Date('1999/01/01'), as.Date('2000/01/01'), by="day"), 200))


trial2 <- mutate(trial, difftimer = difftime(dates,dates2),
                        difftimer2 = as.numeric(difftimer))


tbl_summary(trial2 , by = trt
            ,statistic = all_continuous(continuous2 = TRUE) ~
              c("{median} ({p25}, {p75})","{mean} ({sd})"),
            type = list(all_continuous()~'continuous2')
            ) 
```

michaelcurry1123:
Interesting note about classes but don't know if it is a gtsummary issue but something to be aware of if people ask questions. in the example below I append a fake class onto age and date. if you only append the fake class onto age tbl_summary will run fine, as in it seems to ignore the fake class. However, for the date variable if you append the fake class tbl_summary will not run and give a weird error. 


```
trial <- mutate(trial, dates = seq(as.Date("1970-01-01"), as.Date("1970-01-01")+nrow(trial)-1,1),
                dates2 = sample(seq(as.Date('1999/01/01'), as.Date('2000/01/01'), by="day"), 200))

class(trial$age) <- c("numeric","fakeclass")
class(trial$dates) <- c("Date","fakeclass")

tbl_summary(trial)

```

ddsjoberg:
> when you change the `statistic` argument for `all_continuous()` you get an error for dates. Dates print the min and max with the default but in the below example it throws an error. Perhaps dates should not be influenced by all_continuous? or should the user just put in the non-date variables to the statistic call below?

That is a good point. I think for now, I would like to include it in continuous, but I think the error messaging should be more clear since dates can't be summarized with `quantile()`.

ddsjoberg:
> Interesting note about classes but don't know if it is a gtsummary issue but something to be aware of if people ask questions. in the example below I append a fake class onto age and date. if you only append the fake class onto age tbl_summary will run fine, as in it seems to ignore the fake class. However, for the date variable if you append the fake class tbl_summary will not run and give a weird error.

I think you found a dplyr/vctrs error! Do you want to post an issue to GitHub page? @michaelcurry1123 

``` r
library(dplyr)
packageVersion("dplyr")
#> [1] '1.0.4'

# data frame with date column
df_date <- data.frame(dates = rep_len(Sys.Date(), 5))

# this works
df_date %>%
  dplyr::summarise_at(vars(dates), min)
#>        dates
#> 1 2021-02-27

# adding another class to the column, and the summarize still works
class(df_date$dates)
#> [1] "Date"
class(df_date$dates) <- c("not_a_class", "Date")
df_date %>%
  dplyr::summarise_at(vars(dates), min)
#>        dates
#> 1 2021-02-27

# put the fake class at the end, and the min() fn works and summarise fails
class(df_date$dates) <- c("Date", "not_a_class")
min(df_date$dates)
#> [1] "2021-02-27"
df_date %>%
  dplyr::summarise_at(vars(dates), min)
#> Error: Problem with `summarise()` input `dates`.
#> x `vec_cast.Date.Date()` is implemented at C level.
#> This R function is purely indicative and should never be called.
#> i Input `dates` is `.Primitive("min")(dates)`.
```

<sup>Created on 2021-02-27 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

michaelcurry1123:
@ddsjoberg  I will submit to dplyr!

ddsjoberg:
@michaelcurry1123 FYI, updated with better error messaging.
image

michaelcurry1123:
@ddsjoberg  great! I will take a closer look tomorrow!

michaelcurry1123:
Hey so a few more questions/comments with the example below: 

- for dates and difftime classes if you use `add_p()` it omits the p-values with an error message saying wilcox.text cannot handle those classes, this makes sense but should these classes even be passed to wilcox.test? 
- I specify dates as continuous and it does compute the mean and sd of that variable with the date class and returns a value, instead of producing the warning or an error  


```
trial <- mutate(trial, dates = seq(as.Date("1970-01-01"), as.Date("1970-01-01")+nrow(trial)-1,1),
                dates2 = sample(seq(as.Date('1999/01/01'), as.Date('2000/01/01'), by="day"), 200),
                datetime = as.POSIXct(dates2))


trial2 <- mutate(trial, difftimer = difftime(dates,dates2),
                 difftimer2 = as.numeric(difftimer))

trial2$grade <- factor(trial2$grade, levels = c("II","I","III"),
                       ordered = TRUE, labels = c("II","I","III"))


tbl_summary(trial2 , by = trt
            ,statistic = c('age','marker','ttdeath','dates','difftimer','difftimer2') ~
              c("{mean} ({sd})")
            ) %>% 
        add_p()

```

ddsjoberg:
> * for dates and difftime classes if you use `add_p()` it omits the p-values with an error message saying wilcox.text cannot handle those classes, this makes sense but should these classes even be passed to wilcox.test?

Now that we will allow _any_ class, I think it's best to leave it up to the user to pass appropriate classes to the test functions.  We've kindly added error handling to not break the table when they pass an unsupported class to a test.

> * I specify dates as continuous and it does compute the mean and sd of that variable with the date class and returns a value, instead of producing the warning or an error

I think it's ok to fall back on the R defaults....when you calculate the SD of a date, you get a number..it may not make a lot of sense, but you do get one.

What do you think? @michaelcurry1123

michaelcurry1123:
I think it is fine, people should know what questions they are asking before summarizing data. I am trying to think of other classes to test, are there any big ones that I am not testing? complex numbers?!

ddsjoberg:
I think the most common non-base class that people try to use are the haven_labelled class from the labelled packages.

michaelcurry1123:
ok I checked using the labelled, hmisc, and haven packages to see if the labels affect printing and they do not. They all seem to work well. 



```
labelled::var_label(trial$grade) <- "grade 2"

Hmisc::label(trial$dates) <- "New Date"


trial$trt <- haven::labelled(trial$trt,
                             label = "Treatment 3")

tbl_summary(trial,
            by = "death")

```

ddsjoberg:
@michaelcurry1123 are we set to merge then?

larmarange:
Variable labels does not change variable class.

`haven_labelled` class is applied only when value labels are added to the variable. To be noted that it doesn't change the type of the data (numeric or character), cf. `typeof()`.

A character labelled vector should be seen as a categorical variable.

For a numeric labelled vector, it is a little more complex because we cannot "guess" the true nature of the variable: categorical or continuous. For R, it is still a numeric vector and treated as well. Numeric labelled vectors are often used to store categorical variables but in some cases they could be used to flag certain values in a continuous variable (e.g. to indicate for an age variable that "90" means in fact "90 or +").

Therefore, there are several available options.

The simplest is to consider that a labelled numeric vector is continuous, except if explicitly defined as categorical by users.

A little more complex approach is to autoguess by default, as done by `labelled::unlabelled()`: if all observed values have a label, then categorical, otherwise continuous, of course except if the type of variable is explicitly defined by the user.

Of course, there is also another question: when a labelled vector is defined as categorical, should we display the original codes or the labels.

It should be noticed that labelled vectors (i.e. vectors with value labels) are supposed to be an intermediary format. They should be transformed into factors or into classic numeric vectors before plotting with ggplot2 or before modelling.

When teaching R and the use of labelled vectors, I always say to my students that the good practice is to perform that transformation even before using `tbl_summary`, for example with `labelled::unlabelled()`.

If we follow that philosophy in `gtsummary`, labelled numeric vectors should be considered as continuous and values (and not labels) should be used by `tbl_summary()`. **Eventually**, it could be wise to display a message to say that an `haven_labelled` variable was detected, treated as continuous, and telling the user to use `labelled::to_factor()`, `unclass()`  or `labelled::unlabelled()` to remove that message and treat this variable as categorical if needed. You could also add a link to https://larmarange.github.io/labelled/articles/intro_labelled.html#conditionnal-conversion-to-factors-1

michaelcurry1123:
> It should be noticed that labelled vectors (i.e. vectors with value labels) are supposed to be an intermediary format. They should be transformed into factors or into classic numeric vectors before plotting with ggplot2 or before modelling.

@larmarange @ddsjoberg  With this in mind, that haven_labelled is supposed to be an intermediary format, I do like your suggestion of printing a warning when it sees this format and notify the users to check their data before using gtsummary. It seems that is what ggplot2 does so it makes sense to follow a similar pattern.

ddsjoberg:
Thanks @larmarange @michaelcurry1123 , I agree the message is a good way forward.

@larmarange do users also label survey data like this? (Thinking where to put the message...)

larmarange:
> @larmarange do users also label survey data like this? (Thinking where to put the message...)

The issue is that there is no standard. Value labels are very permissive. You could have observed values with no label and you can add a label to an unobserved value.

Then usage could differ according to data providers or users. If you consider data providers such as national statistic offices, they often label their datasets using value labels but practices could differ from one institution to the other. Regarding  users... well everyone can do how he feels.

Regarding the fact that haven_labelled has been designed as an intermediary class, it is clear from https://haven.tidyverse.org/articles/semantics.html

> The goal of haven is not to provide a labelled vector that you can use everywhere in your analysis. The goal is to provide an intermediate datastructure that you can convert into a regular R data frame. You can do this by either converting to a factor or stripping the labels.

ddsjoberg:
hey hey @larmarange @michaelcurry1123 !

@larmarange I've added an issue to add messaging regarding the haven_labelled class.  We can work that out in a separate pull request.

@michaelcurry1123 I think everything else associated with this PR is complete?

larmarange:
Thanks

michaelcurry1123 (approved):
thanks for addressing the stuff with haven_labelled and other date issues. Looks good!

PR #887, opened 2021-05-03
compatibility of tbl_svysummary() with srvyr

### LINKED ISSUE #886: Compatability between tbl_svysummary() and srvyr::as_survey()-generated tbl_svy objects
**Problem**
Hello there, I've been trying to use @larmarange's brilliant `tbl_svysummary()` to summarize survey data that is knit via Rmd to an HTML {gtsummary} output. Colleagues of mine have had success using srvyr-generated tbl_svy objects (i.e. `srvyr::as_survey_design()`) with `tbl_svysummary()`, but my luck doesn't seem to be the same. The issue appears to be in how `as_survey_design()` is creating the x$call element, which interferes with `tbl_svysummary()`'s attempts to subset the dataset, ending in the error: `Error: $ operator is invalid for atomic vectors`. 

I described my problem in detail in this stackoverflow post: [https://stackoverflow.com/questions/67369601/error-operator-is-invalid-for-atomic-vectors-when-using-srvyr-and-tbl-svysumm](url). @ddsjoberg replied (very quickly!) to suggest making a Feature request on github, so here I am. 

**Proposed Solution**
Allow compatability between `srvyr::as_survey_design()`-generated tbl_svy objects and `tbl_svysummary()`. I _think_ that a workaround where `tbl_svysummary()` tried to call the tbl_svy subsets directly (i.e. `x$strata` rather than `x$call$strata`) would solve the problem. But I am certainly not possessed of the programming capability to understand how to make that happen, or even to know why suggesting such a fix might be worthless...

**Alternatives Considered**
I have tried to implement the same code using `survey::svydesign()` but the Rmd code I am using is designed for use with {srvyr} so the changes required are quite intricate. We are running a very large dataset (cannot share for privacy reasons), which may be contributing to the complexity.

**Additional context**
One key issue that I highlighted in my stackoverflow post is that other colleagues running the exact same Rmd code I have and have not had any issues. I mention this again because if there is some way that the srvyr-tbl_svysummary() connection already exists, I don't want you all to have to do more work!

larmarange:
fix #886

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: gtsummary
 Title: Presentation-Ready Data Summary and Analytic Result
     Tables
-Version: 1.4.0.9002
+Version: 1.4.0.9003
 Authors@R: 
     c(person(given = "Daniel D.",
              family = "Sjoberg",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # gtsummary (development version)
 
+* Updated `tbl_svysummary()` to be compatible with {srvyr} package (#886)
+
 * Updated default header when using `tbl_uvregression(x=)` to `"**Outcome**"` (#867)
 
 * The `tbl_summary(by=)` variable is now added to `include=` by default (#871)
--- a/R/tbl_svysummary.R
+++ b/R/tbl_svysummary.R
@@ -611,6 +611,12 @@ c_form <- function(left = NULL, right = 1) {
         all.vars(x$call$fpc),
         all.vars(x$call$weights)
       )))
+  } else if (!is.null(attr(x, "survey_vars"))) {
+    # "survey_vars" attribute is used by srvyr
+    x$variables %>%
+      select(-any_of(
+        attr(x, "survey_vars") %>% as.character() %>% unlist()
+      ))
   } else {
     x$variables
   }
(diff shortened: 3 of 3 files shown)


ddsjoberg:
All looks good to me!

``` r
library(gtsummary)

data(api, package = "survey")
strat_design_srvyr <- 
  apistrat %>%
  srvyr::as_survey_design(1, strata = stype, fpc = fpc, weight = pw,
                          variables = c(stype, both, starts_with("api"))) 

strat_design_srvyr %>%
  tbl_svysummary(
    # stratify summary statistics by the "both" column
    by = both, 
    # summarize a subset of the columns
    include = c(api00, api99, both),
    # adding labels to table
    label = list(api00 ~ "API in 2000",
                 api99 ~ "API in 1999")
  ) %>%
  add_p() %>%   # comparing values by "both" column
  add_overall() %>%
  as_kable() # convert to kable to display on github
```

| **Characteristic** | **Overall**, N = 6,194 | **No**, N = 2,236 | **Yes**, N = 3,958 | **p-value** |
|:-------------------|:-----------------------|:------------------|:-------------------|:------------|
| API in 2000        | 667 (562, 755)         | 641 (528, 743)    | 671 (588, 765)     | 0.035       |
| API in 1999        | 631 (525, 727)         | 636 (524, 724)    | 625 (526, 727)     | 0.8         |

<sup>Created on 2021-05-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

ddsjoberg (approved):

PR #892, opened 2021-05-04
tidymodels compatibility with `tbl_regression()`

### LINKED ISSUE #647: Is gtsummary compatible with tidymodels and workflows?
@zabore have you ever tried?

### LINKED ISSUE #890: tidymodels and gtsummary
When I use tidymodels it creates all kinds of funky variable names for the covariates in the model, is there a way to manipulate the term cell values to use labelled or something to be able to change the variables names into a more human readable value?

i would think this would need to be post table creation since the input is a fit object that has no way to easily change the names of the terms in the model for tblsummary to use.

or am i just using the package wrong :)

ddsjoberg:
- Added a method for models created with {tidymodels}, `tbl_regression.model_fit()`. The method merely extracts the model fit, `x$fit` and passes the fit to the `tbl_regression()` method. (#890, #647)

closes #890
closes #647 

The current implementation supports models created with parsnip package. The model fit object (`x$fit`) is extracted from the parsnip object, then passed back to `tbl_regression()`.
workflows objects are not directly supported, but when it is passed to `tbl_regression()` they see an error message explaining how modify their code, e.g. `workflows::pull_workflow_fit(x) %>% tbl_regression()`

Checklist for PR reviewer

- [x] Code coverage is suitable for any new functions/features. Review coverage with `covr::report()`. Before you run, set `Sys.setenv(NOT_CRAN="true")` and begin in a fresh R session without any packages loaded. 

- [x] Update NEWS.md with the changes from this pull request under the heading "`# gtsummary (development version)`". If there is an issue associated with the pull request, reference it in parantheses at the end update (see `NEWS.md` for examples).
- [x] Increment the version number using `usethis::use_version(which = "dev")` 
- [x] Run `codemetar::write_codemeta()`
- [x] Run `usethis::use_spell_check()` again
- [ ] Approve Pull Request


R/tbl_regression_methods.R
```diff
@@ -13,6 +13,21 @@
...
 
+#' @export
+#' @rdname tbl_regression_methods
+tbl_regression.model_fit <- function(x, ...) {
+  tbl_regression(x = x$fit, ...)
+}
+
+#' @export
+#' @rdname tbl_regression_methods
+tbl_regression.workflow <- function(x, ...) {
```
zabore, line 24: The only thing I noticed here is that when you use a workflow and get the error message and then follow the instructions to update the syntax, the results of tbl_regression() do not include variable labels that were included in the original data.

zabore (approved):
I tested this with a variety of parsnip models and also workflows and it worked as expected. I left one general comment about the output you get when you do use correct syntax to pass a workflow object to tbl_regression, but I don't know if this is something you can/want to address at this time.

R/tbl_regression_methods.R
ddsjoberg: @zabore thanks for taking a look!

I _think_ the labels are likely lost in the preprocessing that happens in the workflow....hmmm, I wonder if there is a way to get at the labels.....

ddsjoberg:
@zabore couple of questions....

1. When someone passes a workflow object we print a message to first extract the model fit and pass it to `tbl_regression()`. We can easily just call this function for the user in the background. It would require us to add the workflows package as a dependency (can be a suggested package, and not installed by default). 
    - It's easier for the majority of users to just call extract the model fit for them
    - It can be confusing if users are writing their own tidy functions that they may expect to be executed on the workflows object instead of the model fit.
1. The same issue exists for the models created with parsnip, except that we are extracting the model fit for the user. Should the experience for parsnip and workflows be the same? The user needs to extract the model fit for both types and pass to `tbl_regression()`, OR we handle the extraction for both?

zabore:
Good questions @ddsjoberg. I do think it makes sense to have models from {parsnip} and from {workflows} handled in the same way: either by instructing the user how to appropriately change the syntax or by doing it in the background for them. The latter is easier for the user but the former is easier to support.

ddsjoberg:
@zabore ok, I'll think I'll update the workflows method to auto extract the model fit. I'll try to add some messaging in case the parsnip model can't be found.

zabore:
@ddsjoberg Sounds great, I'll review again after.

PR #970, opened 2021-08-25
Adding SMD to `add_difference()`

### LINKED ISSUE #966: Feature request: SMD for categorical variables
Need feature to implement SMD for categorical variables.

At present add_difference (list (all_continous()~ "cohens_d")) works well for continuous variables. 

However, this is not yet available for categorical variables. 

An example of how this is implemented in SAS can be found in the link here: 
https://support.sas.com/resources/papers/proceedings12/335-2012.pdf

This would be extremely helpful to complete the summary tables as many of the variables are categorical in life sciences.

ddsjoberg:
* Added Standardized Mean Difference method to `add_difference()`, wrapping the {smd} package's calculations. (#966)

closes #966

Checklist for PR reviewer

- [ ] Approve Pull Request


malcolmbarrett:
Will it be possible to get SMDs given a set of weights, e.g. propensity weights? It would be great to have a `weights` option, particularly to avoid {survey} when I'm not actually working on a survey problem

ddsjoberg:
> Will it be possible to get SMDs given a set of weights, e.g. propensity weights? It would be great to have a `weights` option, particularly to avoid {survey} when I'm not actually working on a survey problem

Hey @malcolmbarrett ! We don't have an internal mechanism to pass columns that indicate weights outside of the survey package's functionality. If you want to avoid using a survey object, you can define a custom function and pass it to `add_stat()`. The custom function would need to know the weighted column's name. 

I don't use SMD in my work, and my motivation to implement all the features for SMD is sadly not high (😨). But if you have ideas on how to make these features better, please let me know and we can outline the needed pull request together.


```r
library(tidyverse)
library(gtsummary)

gts_smd_w <- function(data, variable, by, ...) {
  smd::smd(x = data[[variable]],
           g = data[[by]],
           w = data[["weights"]], # assuming the weights column is called 'weights'
           na.rm = TRUE) %>%
    select(estimate)
}

trial %>%
  select(trt, age) %>%
  mutate(weights = runif(n())) %>%
  tbl_summary(
    by = trt,
    missing = "no",
    include = -weights
  ) %>%
  add_stat(fns = everything() ~ gts_smd_w) %>%
  modify_header(estimate ~ "**wSMD**") %>%
  modify_footnote(estimate ~ "Weighted Standardized Mean Difference")
```
image

ddsjoberg:
@malcolmbarrett perhaps we can simply add a weight argument to add_difference()? That would be simple enough....

malcolmbarrett:
> perhaps we can simply add a weight argument to add_difference()?

That's exactly the sort of solution I had in mind!

malcolmbarrett:
(cc @lucymcgowan)

ddsjoberg:
@malcolmbarrett will it be a problem that the summary statistics in `tbl_summary()` are not weighted, but the SMD will be weighted?

malcolmbarrett:
I don't think so. What do you think, @LucyMcGowan?

malcolmbarrett:
Is it easy to add an auto-footnote that says the SMD is weighted?

ddsjoberg:
> Is it easy to add an auto-footnote that says the SMD is weighted?

@malcolmbarrett  I added the functionality for weighted statistics. I am trying to think of situations where we can validly calculate a weighted difference in means, and the stratified means don't need to be weighted...perhaps when the weighting doesn't matter within stratum (aka the by variable), but a difference between groups needs the weighting?

``` r
library(gtsummary)
#> #BlackLivesMatter
set.seed(1123)

tbl <-
  trial %>%
  select(age, trt) %>%
  mutate(my_weights = runif(dplyr::n())) %>%
  tbl_summary(
    by = trt, 
    missing = "no",
    include = -my_weights
  ) %>%
  add_difference(all_continuous() ~ "smd",
                 weight = my_weights)
```
image

<sup>Created on 2021-09-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

malcolmbarrett:
Yeah, I see what you mean now that I see the resulting table. A situation where I might want to present the numbers here is showing the baseline differences and indicating if the weights we used, in fact, minimized the SMD. But I would probably explicitly label the column as a weighted SMD and not leave it to a footnote just so it's not confusing that the "difference" is so small. @LucyMcGowan might have some other thoughts on this, as well

malcolmbarrett:
In fact, in practice, I might add two difference columns: the raw difference and the weighted SMD

malcolmbarrett:
I also wonder if @tgerke has any opinion on this

malcolmbarrett:
Sorry to comment on this one thought at a time 😆 

Were I to want weighted means instead, I assume by this there's not currently a way to do that? E.g. {weighted.mean} wouldn't work out of box because there's not a way for it to know where the wts are?

ddsjoberg:
> Were I to want weighted means instead, I assume by this there's not currently a way to do that? E.g. {weighted.mean} wouldn't work out of box because there's not a way for it to know where the wts are?

Exactly, all weighted calculations are done with the survey package, and the weights are recorded inside the survey object.

malcolmbarrett:
Hm then maybe I should get over my concern with using survey (which is more pedagogical, since it's confusing to have to use a package called survey for non-survey purposes) for more complete weighted tables. I still think it's useful to have this simple weighted SMD option, though!

ddsjoberg:
> Hm then maybe I should get over my concern with using survey (which is more pedagogical, since it's confusing to have to use a package called survey for non-survey purposes) for more complete weighted tables. I still think it's useful to have this simple weighted SMD option, though!

I think the primary way weights are used, outside of actual survey analysis, are perhaps weighted propensity analyses? Maybe we can add an example of propensity analysis using gtsummary to the table gallery?

I am unsure about keeping the weight argument in `add_difference()` because I can't think of a scenario where unweighted summary statistics are valid when weighting is needed to calculate a difference? 😕

malcolmbarrett:
> Maybe we can add an example of propensity analysis using gtsummary to the table gallery?

I think this would be great!

>  I can't think of a scenario where unweighted summary statistics are valid when weighting is needed to calculate a difference?

I get this, but IMO in propensity score weighting, they're still valid because they're still the true means. The weighted difference is then not the corrected difference per se, as you might consider it in survey design, but an alternative

malcolmbarrett:
But @LucyMcGowan and I just spoke a bit and agreed that, in practice, fully-weighted tables are a bit more useful

ddsjoberg:
> But @LucyMcGowan and I just spoke a bit and agreed that, in practice, fully-weighted tables are a bit more useful

@malcolmbarrett the tableone package has a great vignette on using the package with propensity scores (https://cran.r-project.org/web/packages/tableone/vignettes/smd.html). They also create survey object to prepare the summary stats. I will need to look into how they extract the weights for the SMD calculation....we don't currently export a `add_difference.tbl_svysummary()` method!

malcolmbarrett:
Yes, we've used tableone in the past but want to switch to gtsummary since that's what I use pretty exclusively now. So using that package to help design a good way in gtsummary will likely meet our needs!

ddsjoberg:
hey hey @malcolmbarrett ! 

FYI, I merged this pull request without the `add_difference(weight=)` argument. I didn't want to add another argument without giving it more thought first. If you have examples where unweighted summary statistics are valid and we need a weighted standardized mean difference, please do post an issue and we can look into further!

malcolmbarrett:
Thanks for the update! The only use-case I have is when I might want the non-weighted SMD and weighted SMD side-by-side, e.g. a table with the unweighted summary statistics with the SMD then a propensity-score weighted SMD, the idea being that you're demonstrating what it looks like post-weighting in addition to the raw data

PR #1039, opened 2021-12-01
Better error messages for `stat_fns` in `tbl_custom_summary()`

larmarange:
- If just a function is passed, it is now transformed into `everything() ~ fun`
- Check if it is a formula or a list of formulas


ddsjoberg:
@larmarange this looks good!

Before we merge, can you add a unit test for the new functionality?

larmarange:
Two additional tests have been added.

R/tbl_custom_summary.R
```diff
@@ -284,7 +284,20 @@ tbl_custom_summary <- function(
...
+  }
+  if (rlang::is_bare_formula(stat_fns))
+    stat_fns <- list(stat_fns)
+  if (!rlang::is_bare_list(stat_fns))
+    stop(paste0(
+      "'stat-fns' argument must be a formula or a list of formulas.\n",
+      "LHS of the formula is the variable specification, ",
+      "and the RHS is a function:\n",
+      "everything() ~ continuous_summary(\"var\")"
+    ), call. = FALSE)
```
ddsjoberg, line 299: @larmarange 

The formula list will eventually be processed by `broom.helpers::.formula_list_to_named_list()`. Rather than adding these checks and messaging here, would it not be preferable to make the messaging better in that function instead?

One other thing to consider...I've made a syntax error with these lists many many times when I want to apply the same instructions to all variables. Should we add an input type to `broom.helpers::.formula_list_to_named_list(typeof = NULL)`. If a typeof argument is supplied we could do two things: 
1. If we had `broom.helpers::.formula_list_to_named_list(x, typeof = rlang::is_function)` and `rlang::is_function(x)` is TRUE, we auto convert this to `list(everything() ~ x)`.
2. Add a check that each element on the RHS of the formula satisfies `rlang::is_function(.)` 

What do you think?

larmarange: @ddsjoberg  Yes it could be a good idea to improve messaging in `broom.helpers::.formula_list_to_named_list()` with the possibility of indicating a certain type.

To adapt the error message, we should have a way to indicate what is expected on the RHS side, i.e. a number, a function, a text... Maybe having an option in `broom.helpers::.formula_list_to_named_list()` to pass a custom string that will be glued in the error message? or to adapt the examples indicated in the error message? In that case, it should be possible to make the error message more custom in other parts of `gtsummary`.

Regarding point 2, I do not know if it is feasible. Right now you can provide the name of a function, e.g. `"mean"` and it is working, but not sure where it is evaluated as a function. We just need to be sure that the element is already evaluated at that step.

Would you like to propose a PR in `broom.helpers`? You are more familiar than me with `.formula_list_to_named_list()` and with possible implications for `gtsummary`.

Thanks

ddsjoberg:
Hmmmmm @larmarange .... I started a pull request to implement as discussed. But there is a potential issue. There are some arguments that accept lists on the RHS. For example, `tbl_summary(digits = all_continuous() ~ list(style_number, style_percent))`.

If we implement a feature that infers `everything() ~ x` when x is neither a formula nor list (e.g. `x = 1` --> `x = everything() ~ 1`), we'll have inconsistent syntx,

As an example, we would want `tbl_summary(digits = list(style_number, style_percent))` to be inferred as `tbl_summary(digits = everything() ~ list(style_number, style_percent))` .... but there is no way to infer this, right?

Regardless, error messaging can be improved, which i'll continue working on that.

larmarange:
> If we implement a feature that infers `everything() ~ x` when x is neither a formula nor list (e.g. `x = 1` --> `x = everything() ~ 1`), we'll have inconsistent syntx,
> 
> As an example, we would want `tbl_summary(digits = list(style_number, style_percent))` to be inferred as `tbl_summary(digits = everything() ~ list(style_number, style_percent))` .... but there is no way to infer this, right?
>

Right. It should not be part of `.formula_list_to_named_list()`. 

Eventually, is some functions (e.g. like in `tbl_custom_summary()`) we could do a correction with a warning indicating that this is not correct syntax and providing the correct syntax. But it will remained embed within relevant function and not a generic element handled by `.formula_list_to_named_list()`.

> Regardless, error messaging can be improved, which i'll continue working on that.

Yes, it could be handled by `.formula_list_to_named_list()`, having an option to improve error messages.

larmarange:
See https://github.com/ddsjoberg/gtsummary/pull/1039/commits/835f1f31d5e4a17e866e348a05fefbcee090aaf4 as an example of a specific warning, handled directly by the function and not as generic

R/tbl_custom_summary.R
```diff
@@ -284,7 +282,16 @@ tbl_custom_summary <- function(
   tbl_custom_summary_inputs <- as.list(environment())
 
   # checking function inputs ---------------------------------------------------
-  # verify if checks about stat_fns should be added at this stage
+  if (rlang::is_function(stat_fns)) {
```
ddsjoberg, line 285: This message allows inconsistent syntax to the other arguments in the package. I would prefer to keep the helpful error messaging we just added in broom.helpers. The user will still be prompted to update their syntax.

ddsjoberg (changes requested):

R/tbl_custom_summary.R
```diff
@@ -427,7 +434,9 @@ generate_metadata_custom_summary <- function(data, stat_fns, include,
       x = stat_fns,
       data = data %>% select(any_of(include)),
       var_info = meta_data_to_var_info(meta_data),
-      arg_name = "stat_fns"
+      arg_name = "stat_fns",
+      type_check = .is_convertible_as_function,
```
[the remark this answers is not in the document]
larmarange: :-)

R/tbl_custom_summary.R
larmarange: It has been changed

ddsjoberg (approved):

larmarange:
thanks

PR #1108, opened 2022-01-14
Add bold/italic cells for `as_kable_extra()` output

### LINKED ISSUE #1107: Feature request: Update `as_kable_extra()` to bold cells in table.

### LINKED ISSUE #1114: Should we add `kable(align=)` argument?
Then all alignment specifications from gtsummary will also appear in the resulting kable tables

ddsjoberg:

* The output from `as_kable_extra()` may now include bold and italic styling. User must specify the output type to take advantage of this feature, e.g. `as_kable_extra(format = "latex")` 

closes #1107 
closes #1114 

Checklist for PR reviewer

- [ ] Approve Pull Request


ddsjoberg:
@shannonpileggi if you have a chance to take a look at this that would be fantastic! I am not too familiar with kableExtra, unfortunately.

- The cells that are being bold or italic are modified using `kableExtra::cell_spec(., format=, bold=, italic=)`, where format is being pulled from `as_kable_extra(format=)` if present. If user does not specify `format=`, bolding and italics are skipped
- There seems to be issues with the latex or HTML code not rendering and appearing verbatim in the cells. I am not sure why.
- Anything passed in the dots of `as_kable_extra(...)` is added to the `knitr::kable(...)` call. Users can pass other needed arguments using the dots, e.g. `as_kable_extra(format = "latex", escape = FALSE)` 

If this is something you think you can tackle in the next week or so, we can add it to the next gtsummary release. Otherwise, we can add it in the following release.

shannonpileggi:
This is tricky and I don't have any immediate ideas on how to fix, but I do see a problem.

```
---
title: "Untitled"
author: "Daniel D. Sjoberg"
date: "1/14/2022"
output: 
  pdf_document:
        keep_tex: yes
---
```


```{r cars}
library(gtsummary)

tbl <-
  trial %>%
  select(age, response) %>%
  tbl_summary() %>%
  bold_labels()

tbl %>% as_kable_extra(format = "latex")

```

If you run this .Rmd with the `keep_tex: yes` option then you can inspect the tex code that gets generated.

Here, you can see that `\textbf{}` is being translated as `\textbackslash{}textbf`

```
\begin{tabular}{l|l}
\hline
Characteristic & N = 200\\
\hline
\textbackslash{}textbf\{Age\} & 47 (38, 57)\\
\hline
\hspace{1em}Unknown & 11\\
\hline
\textbackslash{}textbf\{Tumor Response\} & 61 (32\%)\\
\hline
\hspace{1em}Unknown & 7\\
\hline
\multicolumn{2}{l}{\rule{0pt}{1em}\textsuperscript{1} Median (IQR); n (\%)}\\
\end{tabular}
```

so I wonder if there needs to be some extra escaping on the `\textbf`.

ddsjoberg:
I tried HTML output, and included these arguments `as_kable_extra(tbl, format = "html", escape = FALSE)` and it worked perfectly.

I then tried PDF output and had errors. I did some reading on the kableExtra vignettes, and found this
image

And this

image


I did some prelim testing using `column_spec()` instead of `cell_spec()` and it seemed to work pretty well. I'll updated and we can see where that gets us

ddsjoberg:
OMG Dr @shannonpileggi , I think using the `column_spec()` function instead of `cell_spec()` worked! NO need to specify `format=`, `escape=` or anything. So nice!

Can you review this PR and if it's working for you as well, we can add to this weeks gtsummary release!

shannonpileggi:
Awesome!  I took it for a whirl and it is so nice.  

# A great first run!

I tried to replicate my more complex specs exactly, and I got almost there!

```
tbl <- penguins %>% 
  tbl_summary(
    by = species
  ) %>% 
  bold_labels() 

tbl %>% 
  as_kable_extra(
    format = "latex",
    booktabs = TRUE,
    longtable = TRUE,
    linesep = ""
  ) %>%
  kable_styling(
      position = "left",
      latex_options = c("striped", "repeat_header"),
      stripe_color = "gray!15"
    )
```

😍
image

# More complex is harder

However, when I tried to insert line breaks in table headers I couldn't figure it out.  😟

```
tbl_mod <- tbl %>% 
  modify_header(
    all_stat_cols() ~ "{level}\nN = {n}"
  )   

tbl_names <- tbl_mod$table_styling$header %>% 
  filter(str_detect(column, "stat")) %>% 
  pull(label)

tbl_mod %>% 
  as_kable_extra(
    format = "latex",
    booktabs = TRUE,
    longtable = TRUE,
    linesep = "",
    col.names = linebreak(tbl_names, align = "l"),
    escape = FALSE
  ) 
```
With or without the escape option, this yields the error

```
Error in knitr::kable(., col.names = c("Characteristic", "Adelie\nN = 152", : 
formal argument "col.names" matched by multiple actual arguments
```
and the traceback shows that it ends on `column_spec`  😱

image

Do you know how users could insert line breaks into headers with this additional functionality?

# Escaping

Also, knitting the first code  example with `escape = FALSE` fails.

The not-knitted pdf has this error:

image

because now any special latex characters, like  `%`, need to be changed to `\%`.  If the user wanted the option to use `escape`  to create manual line breaks, could that be implemented? Or maybe you could borrow from the `linebreak` function to translate any tables containing `\n` to execute the line breaks?

ddsjoberg:
@shannonpileggi 

Delighted the basic case is working out!

# `kable(col.names=)`

When you're trying to specify the line breaks in the column headers, you're passing the `col.names=` argument to `kable()`. But this arg is already being used to define the column headers. The first error you're getting is about duplicate calls to the `col.names=` argument. So this would have never worked with or without the bolding in `column_spec()`. But what to do about it?
    - We could wrap the column headers in the `linebreak()` function by default so there is no need to specify it yourself. An issue would be that you cannot change the linebreak symbol.
    - We could be more careful about arguments passed. In the help file, be clear about what args are in use, and if a user passes the same arg, we simply use their version (rather than keeping both, which results in an error).
    
# Escaping

I think we'll need to have a conversation about escaping, what special characters would need it, etc.  I guess we could add an arg about escaping special characters....after all cells have been formatted we could replace all percent signs with a slash in front. But idk, I don't like the idea of changing things this way. Maybe it's better to have a vignette showing power-users how to take advantage of all features. One solution is that when they create the original tables they need to request that the statistics are `"{p}\%"`, adding the escape character themselves?

For these complex issues, my instinct is to defer to the user to handle it themselves...there is too much about it that I don't understand! But maybe you can help me understand!

shannonpileggi:
# `kable(col.names=)`
> We could wrap the column headers in the `linebreak()` function by default so there is no need to specify it yourself. An issue would be that you cannot change the linebreak symbol. 

You can pass multiple `linebreakers`, but it does not result in expected behavior.  Note that two vectors are returned instead of single vector. 

``` r
x <- "A \n B <br> C"
kableExtra::linebreak(x)
#> [1] "\\makecell[l]{A \\\\ B <br> C}"
kableExtra::linebreak(x, linebreaker = "<br>")
#> [1] "\\makecell[l]{A \n B \\\\ C}"
kableExtra::linebreak(x, linebreaker = c("\n", "<br>"))
#> [1] "\\makecell[l]{A \\\\ B <br> C}" "\\makecell[c]{A \n B \\\\ C}"
```

Another thing that might make this tricky is the `align` argument in `linebreak`.

`linebreak(x, align = c("l", "c", "r"), double_escape = F, linebreaker = "\n")`

As this was something I struggled with previously. When you use `linebreak` and don't specify `align` I believe it cycles through the options, as you can see from the tex output above.

In addition, linebreaks might be something that the user would want to control for both column names and labels If this could be worked out.

> If a user passes the same arg, we simply use their version (rather than keeping both, which results in an error).

This is also a great solution!  We could experiment with getting labels to work similarly.
 

# Escaping

> One solution is that when they create the original tables they need to request that the statistics are `"{p}\%"`, adding the escape character themselves?

I think this is a great solution.  It might take a bit of practice to iron out, but that can be left for for the user.  Here are my first tries.

```
tbl <-
  trial %>%
  select(age, response) %>%
  tbl_summary(
    statistic = list(all_categorical() ~ "{p}\% ({n} / {N})")
  ) 
```
image

then I tried 

```
tbl <-
  trial %>%
  select(age, response) %>%
  tbl_summary(
    statistic = list(all_categorical() ~ "{p}\\% ({n} / {N})")
  ) %>%
  bold_labels()

tbl %>% 
  as_kable_extra(
    format = "latex",
    escape = FALSE
  )
```

and there was no error! 🎉

image

The footnote needs to be modified as well, but the user can figure that out if they have made it this far.

ddsjoberg:
@shannonpileggi awesome! you're so good!

# linebreaks

For now, I'll just update the construction of the kable call that will allow users to write over the default. Something like this:

```r
# constructs call to kable, and allows users to overwrite default arg values.
.construct_call_to_kable <- functon(x, ...) {
  dots <- rlang::dots_list(...)
  kable_args <-
    # default args
    list(caption = x$table_styling$caption,
         col.names = dplyr::filter(x$table_styling$header, .data$hide == FALSE)$label) %>%
    # update with any args user passed
    purrr::list_modify(!!!dots) %>%
    purrr::compact()

  expr(knitr::kable(!!!kable_args))
}
```

The downside is that if lose the column header specification and users will need to do in manually. The up side is that users can customize however they need, which I think is better overall.

# escaping

OK, let's put it on the user to add the escape characters (which they may have already been doing!), and it wouldn't be a breaking change. I agree, if the user is good enough to add escapes to special chars, they can fix the footnote at the end.

Do you want to add a pro-tips section to the `as_kable_extra()` illustrating how to use these various features?

shannonpileggi:
Awesome - I'll give some testing when updates are implemented.

And yes, happy to add pro-tips, and of course a blog post to follow!

ddsjoberg:
@shannonpileggi pushed the updates! Can you check this allows for everything you'd need and add that Tips section to the `as_kable_extra()` help file?

shannonpileggi:
Hey!  I'm still tinkering, but while I am at it I had another thought.  

```
tbl <-
  trial %>%
  select(age, response) %>%
  tbl_summary() %>%
  bold_labels()

tbl
````
Returns 
image

in the vewer.

Then with 

```
as_kable_extra(tbl)
```

you get 

image


and with 

```
as_kable_extra(tbl, format = "latex")
```

you get 

image

is it worth it to try to get the headers to follow consistent formatting (bolding) with the .html viewer, or should that be left up to more advanced users?

**Comment by ddsjoberg** [comment]
@shannonpileggi yeah, that would be great! But I don't know how to do it 😆 

The header strings are already saved with the markdown syntax (double stars). Can we get the headers to use the markdown syntax? If not, how do you bold partial text in a kable header?

**Comment by shannonpileggi** [comment]
Ok, I tried a few things.  `row_spec` isn't great because it will make the whole row bold, rather than elements.

Otherwise, looks like `escape` creates some tricky problems.


```
tbl <-
  trial %>%
  select(trt, age, response) %>%
  tbl_summary(
    by = trt
  ) %>%
  bold_labels()
```

Basic manual specification of `col.names` works:

```
names_1 <- c("AAA", "BBB", "CCC")
tbl %>% 
  as_kable_extra(
    format = "latex",
    col.names = names_1
  )
```
```
\begin{tabular}{>{}l|l|l}
\hline
AAA & BBB & CCC\\
\hline
\textbf{Age} & 46 (37, 59) & 48 (39, 56)\\
\hline
\hspace{1em}Unknown & 7 & 4\\
\hline
\textbf{Tumor Response} & 28 (29\%) & 33 (34\%)\\
\hline
\hspace{1em}Unknown & 3 & 4\\
\hline
\multicolumn{3}{l}{\rule{0pt}{1em}\textsuperscript{1} Median (IQR); n (\%)}\\
\end{tabular}
```

image


However, when you need special formatting in the column names, I can't figure out how to get the tex to correctly to render with `escape = TRUE`, I have only been successful `escape = FALSE`.

```
names_2 <- c("\\textbf{A}AA", "\\textbf{B}BB", "\\textbf{C}CC")
tbl %>% 
  as_kable_extra(
    format = "latex",
    col.names = names_2,
    escape = FALSE
  )
```

```
\begin{tabular}{>{}l|l|l}
\hline
\textbf{A}AA & \textbf{B}BB & \textbf{C}CC\\
\hline
\textbf{Age} & 46 (37, 59) & 48 (39, 56)\\
\hline
\hspace{1em}Unknown & 7 & 4\\
\hline
\textbf{Tumor Response} & 28 (29%) & 33 (34%)\\
\hline
\hspace{1em}Unknown & 3 & 4\\
\hline
\multicolumn{3}{l}{\rule{0pt}{1em}\textsuperscript{1} Median (IQR); n (\%)}\\
\end{tabular}
```


BUT, even though the tex correctly renders, the pdf won't, and I'm not quite sure why.

image

In general, it looks like `escape = FALSE` is not yet compatible with `as_kable_extra`.

This alone yield the same error.

```
tbl %>% 
  as_kable_extra(
    format = "latex",
    escape = FALSE
  )
```

I don't know how to render the bold without `escape`, and I don't know how hard it would be to escape-proof `as_kable_extra`.


And I did test that it works without `as_kable_extra`.

```
tbl <-
  trial %>%
  select(trt, age, response) %>%
  tbl_summary(
    by = trt,
    statistic = all_categorical() ~ "{n}"
  ) 

names_2 <- c("\\textbf{A}AA", "\\textbf{B}BB", "\\textbf{C}CC")

tbl %>% 
 as_tibble() %>% 
   kbl(
    format = "latex",
    col.names = names_2,
    escape = FALSE
  )
```

yields

image

when knitted to pdf.

There source code of kableExtra might be some help for a few things if we can get escaping to work.


Digging into `kableExtra::cell_spec` [here](https://github.com/haozhu233/kableExtra/blob/9e459cd0c08a3070cc32684fe3ee6f9bafe84949/R/cell_spec.R#L169), you might be able to borrow a bit of their code.


``` r
bold <- TRUE
x <-"Abc"
# if (escape) x <- escape_latex(x)
x <- sprintf(ifelse(bold, "\\textbf{%s}", "%s"), x)
x
#> [1] "\\textbf{Abc}"
```

and there is also a hidden `kableExtra:::escape_latex()` function, but I'm not sure how that fits in with the knitting process - I did not get a result to render correctly with it.

``` r
names_2 <- c("\\textbf{A}AA", "\\textbf{B}BB", "\\textbf{C}CC") %>% 
kableExtra:::escape_latex()
names_2
[1] "\\textbackslash{}textbf\\{A\\}AA" "\\textbackslash{}textbf\\{B\\}BB"
[3] "\\textbackslash{}textbf\\{C\\}CC"
```



<sup>Created on 2022-01-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

ddsjoberg:
OH goodness, this is giving me flashbacks why support in `as_kable_extra()` is so basic 😆 It's so tricky!

Here's what I am gathering are the issues.
1. Using `column_spec()` for bold/italic works great for _most_ output (and using all the defaults), but fails if user specifies `escape = FALSE`.
2. You _need_ `escape = FALSE` to bold the headers, and therefore, we have incompatible way of bolding.
3. We know that the output using `cell_spec()` works with `escape = FALSE` (requires it even). But `cell_spec()` can't be the default, because I think it could cause issues with other users if we changed the `escape=` default value.

I see a two primary ways forward.
1. Do nothing and leave it to users to add whatever custom code they need for their particular situation 🤷🏼 
2. Add an argument to `as_kable_extra(spec_method = c("none", "column", "cell"))` for the users would specify the method used to bold/italicize cells in a table.
    - `"none"` no bold/italic code in the table body will be used (this is what we currently do)
    - `"column"` Use `column_spec()` to bold cells and we'll note this only works with `escape = TRUE` (the default value)
    - `"cell"` Use `cell_spec()` to bold cells and note this only works with `escape = FALSE`

I think we should avoid any solutions that required us to parse/escape anything ourselves. This is just something I am not familiar enough to confidently do, and I am hesitant to incorporate code I don't feel like I can maintain in the future.

What do you think the best way forward is?

ddsjoberg:
@shannonpileggi FYI Number 3 is currently implemented in the pull request branch. I did some minor testing, and seems to work well. But I am pretty sure you can break it somehow lol

shannonpileggi:
Um ok you are a miracle worker!  Looks like *both* line breaks and bolding is working in the tables names *with* `escape = FALSE`. 🎉

It looks like bold formatting applied in gtsummary column names is not transferring to latex column names - so that can be left for the user or attempted in a next version.  What do you think?

Do you want the pro-tips focused on pdf only, or both pdf & html? I am guessing for the pdf ones I should do manual screen shots instead of using `gt.doc.images` - that would be html only, right?

ddsjoberg:
> Um ok you are a miracle worker! Looks like _both_ line breaks and bolding is working in the tables names _with_ `escape = FALSE`. 🎉

I'm just using the codes you wrote!

> It looks like bold formatting applied in gtsummary column names is not transferring to latex column names - so that can be left for the user or attempted in a next version. What do you think?

The column headers in gtsummary are using markdown syntax (e.g. double stars to bold), and kableExtra doesn't support it. There is an argument to `as_kable_extra(strip_md_bold=TRUE)` that removes all the double stars from the column headers. From what I can tell, there is no way to respect the markdown syntax with {kableExtra} headers. I think it _could_ be fun (in the future) to write some vvv fancy regex replacements exchanging anything in double stars with HTML or LaTex bold codes (same for italic, then again for underscore bold/italic syntax, then again for other markdown syntax....ugh, then it becomes exhausting......) 

> Do you want the pro-tips focused on pdf only, or both pdf & html? I am guessing for the pdf ones I should do manual screen shots instead of using `gt.doc.images` - that would be html only, right?

I think PDF ones are the tricky to use, right? I think we'd need to touch the following:
1. How the cells are bold/italic depending on the value of escape.
2. Something more on escape? idk...like, when to use it and not?
3. An example with `linebreak()` would also be nice.
4. Is that everything?

The `gt.doc.images` package only handles gt, gtsummary, and flextable images. I don't know how we'd easily modify it to get images from LaTeX tables? I think we can just make the screenshots ourselves.

What else do you think should go here? I think you have the better intuition for what belongs here!

PR #1113, opened 2022-01-19
Sp kable extra updates

shannonpileggi:
**What changes are proposed in this pull request?**  added documentation and examples for `as_kable_extra()`

Do you think the description and examples cover it? The examples are pretty text heavy - would that discussion be better placed elsewhere?

Can you confirm the examples work for you? And you don't want any modifications? If so, I can add in the images next.

Lastly, in example 2, it is very weird, but I cannot explain to you why
```
#'   modify_footnote(
#'      update = all_stat_cols() ~ "Median (IQR); n (%)"
#'   ) 
```

works in the printed result; but without it, the percent sign gets an additional character.


ddsjoberg:
Notes from call:
1. Migrate comments in examples section to a new section in the help file section of the file. You can start a new section with `#' @section LaTeX Tips:`
2. You mentioned adding images and I think that is great idea. We can't use the gt.doc.images package to create them. We'll have to save them ourselves, and add them to the `~man\figures\` folder. It'll be important that the files you make DO NOT end in `_ex[:digit:].png` or the gt.doc.images could delete it when it's doing a re-fresh of example files.
3. For example 1 output, I like your suggestion of adding examples of both HTML and PDF output. We could do it with something like this:
```r
#' @section Example Output:
#' \if{html}{Example 1, HTML Output}
#' 
#' \if{html}{\figure{as_kable_extra_ex1_html.png}{options: width=50\%}}
#' 
#' \if{html}{Example 1, LaTeX Output}
#' 
#' \if{html}{\figure{as_kable_extra_ex1_latex.png}{options: width=50\%}}
```

shannonpileggi:
For the image previews, I manually snipped the pic of the html output b/c I didn't want to redocument the whole package. Let me know if any of the 4 images need to be resized.

> Oh, and we didn't discuss this, do you think an example bolding the header row would be helpful? Maybe in example 1?

Let's chat about this one

ddsjoberg:
Woohoo! Thank you @shannonpileggi ! It's gorgeous!

PR #1388, opened 2022-11-19
Bug fix in `tbl_custom_summary()`: the full dataset (including missin…

larmarange:
According to the doc, `full_data` is supposed to be the full dataset, including missing observations


ddsjoberg:
great, thanks for the update! Do you mind adding a unit test for this?

larmarange:
> great, thanks for the update! Do you mind adding a unit test for this?

of course, will do

larmarange:
Unit test has been added

PR #1404, opened 2022-12-15
Use `vec_paste0()` helper for better recycling

DavisVaughan:
Hi there, we were running revdeps for dplyr 1.1.0 and this package was indirectly flagged through a breakage in {logitr}, which relies on this package.

``` r
library(logitr)
library(gtsummary)

model <- logitr(
    data    = yogurt,
    outcome = "choice",
    obsID   = "obsID",
    pars    = c("price", "feat", "brand")
)
#> Running model...
#> Done!

model |> 
    tbl_regression()
#> Error in `dplyr::rename_with()` at dplyr/R/join.R:321:2:
#> ! `.fn` must return a vector of length 0, not 1.

#> Backtrace:
#>     ▆
#>  1. ├─gtsummary::tbl_regression(model)
#>  2. ├─gtsummary:::tbl_regression.default(model) at gtsummary/R/tbl_regression.R:108:2
#>  3. │ └─... %>% ... at gtsummary/R/tbl_regression.R:250:2
#>  4. ├─dplyr::relocate(., starts_with("modify_stat_"), .after = last_col())
#>  5. ├─dplyr::full_join(., x$table_styling$header, by = character()) at dplyr/R/relocate.R:50:2
#>  6. ├─dplyr::rename_with(., .fn = ~paste0("modify_stat_", .), .cols = everything()) at dplyr/R/join.R:321:2
#>  7. └─dplyr:::rename_with.data.frame(., .fn = ~paste0("modify_stat_", .), .cols = everything()) at dplyr/R/rename.R:59:2
#>  8.   └─cli::cli_abort("{.arg .fn} must return a vector of length {length(sel)}, not {length(new)}.") at dplyr/R/rename.R:79:4
#>  9.     └─rlang::abort(...)
```

<sup>Created on 2022-12-15 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

`rename_with()` now checks that `fn` returns a result that is the same size as the input, i.e. 3 column names go in so 3 should come out. This also needs to apply when there are 0 column names, but `paste0()` doesn't recycle well when there are size 0 elements mixed with size 1 elements. I've added a `vec_paste0()` helper that we use in situations like this. You already rely on vctrs through dplyr, so this isn't an additional dependency for you.

We plan to submit dplyr 1.1.0 on January 27th.

This should be compatible with both dev and CRAN dplyr. It would help us out if you could go ahead and send a patch version of your package to CRAN ahead of time! Thanks!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -87,7 +87,8 @@ Imports:
     rlang (>= 1.0.3),
     stringr (>= 1.4.0),
     tibble (>= 3.1.6),
-    tidyr (>= 1.1.4)
+    tidyr (>= 1.1.4),
+    vctrs (>= 0.5.1)
 Suggests: 
     aod (>= 1.3.1),
     broom.mixed (>= 0.2.9),
--- a/R/add_ci.R
+++ b/R/add_ci.R
@@ -200,7 +200,7 @@ add_ci.tbl_summary <- function(x,
           dplyr::relocate(all_of(sort(cols_to_order)),
                           .before = all_of(sort(cols_to_order)[1])) %>%
           dplyr::rename_with(
-            .fn = ~paste0(
+            .fn = ~vec_paste0(
               "ci_",
               stringr::str_replace(., pattern = "_ci$", replacement = "")),
             .cols = matches("^stat_\\d+_ci$")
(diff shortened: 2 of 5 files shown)


ddsjoberg:
Thank you @DavisVaughan !

PR #1496, opened 2023-04-17
updating precision in `round2()`

### LINKED ISSUE #1472: Remove `dplyr::storms` from unit testing
We use the `dplyr::storms` in the unit testing. But this dataset was updated in the last dplyr release to include more storms, which resulted in failing tests. These have been updated for now, but we should remove it so future updates don't break us

### LINKED ISSUE #1494: Bug Report: round2 is unreliable
`round2` does not reliably solve the issue of `round` not behaving in the expected way.

```r
round2(2.5)
#> 2
```
It seems to be an issue with `(number + .Machine$double.eps) == number` being true for numbers greater than 2. Perhaps it would make sense to implement `round2` as in this link? https://stackoverflow.com/questions/12688717/round-up-from-5/12688836#12688836 This is what `janitor` references in its implementation of `round_half_up`: https://github.com/sfirke/janitor/blob/main/R/round_half_up.R

ddsjoberg:
* Updating the levels of precision used in `round2()`, which is used in the background for every rounded/formatted number in a gtsummary table. (#1494)

closes #1494
closes #1472

- [ ] Ensure all package dependencies are installed by running `renv::install()`
- [ ] Code coverage is suitable for any new functions/features. Review coverage with `withr::with_envvar(new = c("NOT_CRAN" = "true"), covr::report())`. Begin in a fresh R session without any packages loaded. 

- [ ] Approve Pull Request


nikostr:
Thanks for looking into this! Won't this implementation simply shift where problems occur? Running

```{r}
number <- 10^seq(1,15)
(number + sqrt(.Machine$double.eps)) == number
#> [1] FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE
round2(2.5e10, digits=-10)
#> [1] 2e+10
```

and since `sqrt(.Machine$double.eps` is aboue `1.5e-8` any numbers in that range will start behaving badly.

ddsjoberg:
@nikostr oh yes, i was just in the middle of testing out a few things. I was hoping to land on a solution that had fewer computations....but as you noted, the previous code wasn't perfect.

I've tested the janitor/stackoverflow solution too, and it's better, but has some issue too `round2(0.499999999, digits = 0) = 1` ¯\_(ツ)_/¯

PR #1898, opened 2024-08-12
Fix newline formats when quarto rendering with {gt}

### LINKED ISSUE #1896: Header formatting issue with quarto
Hello

I use R 4.4.1 on Debian testing with pandoc 3.1.3 and quarto 1.1.553
A very simple Quarto  document :

```
---
title: "test"
format: pdf
---

```{r}
library(gtsummary)

trial %>%
  select(age, grade, trt) %>%
 tbl_summary(by = trt)
```
```
The result is correct with an export in html or when I just compile the chunk but in pdf I have this result : 
![tab1](https://github.com/user-attachments/assets/d5521be8-f84f-4a61-b7be-baadd6eb4285). 

There is no error message during compilation. I was reported the same problem on a Shiny compilation from a source Rmarkdown

### PR DESCRIPTION
**Pull request by Melkiades**
Added a call to remove newlines (`\n`) from column labels when using `as_gt()` for {quarto} renderings. Final output from issue example:
image

closes #1896 

- [ ] Ensure all package dependencies are installed: `renv::install()`
- [ ] PR branch has pulled the most recent updates from master branch: `usethis::pr_merge_main()`

- [ ] Approve Pull Request

### CODE REVIEW

**Review by Melkiades — COMMENTED** [review]
@ddsjoberg like now we always remove new lines from header. Do you think we should localize this to quarto rendering somehow? or maybe just avoiding multiple \n?

**Comment by Melkiades** [comment]
> @ddsjoberg like now we always remove new lines from header. Do you think we should localize this to quarto rendering somehow? or maybe just avoiding multiple \n?

The (new) current fix is in the knitr print method and not anymore in as_gt() call stack. In this way, it will only affect knitr (and quarto afaik) rendering where newlines are removed.

Try the following in your quarto code:
```{r}
library(gtsummary)
a <- trial |>
  tbl_summary(
    by = trt
  )
print(a$table_styling$header)
#> [1] "variable"              "var_type"              "row_type"             
#> [4] "var_label"             "**Characteristic**"    "**Drug A**  \nN = 98" 
#> [7] "**Drug B**  \nN = 102"
a
```
You get:
image

If you render it with quarto you get:
image

Finally if you add `a | > as_gt()` at the end of last block of code AND render you get the original miss-format:
image

I think the latter case is happening because that is a bug in {gt} that is not circumvented by our print method in {gtsummary}. Should we add an issue there and + a warning in the fix? The latter will generate a bit of clatter though @ddsjoberg

ddsjoberg:
@Melkiades in the `knit_print()` method, i think we also need to add `knitr::is_latex_output()` as a condition to check so that the line break is only removed for latex output (since it work properly for the other output options).

ddsjoberg:
Thanks @Melkiades ! Do you think there is a way we can add a unit test for this? Also, can we add a note in the help file for `as_gt()` that for PDF output, the `\n` line break is stripped due to a bug? I think we should also add a date, so we could perhaps test it again in the future, say after a year or so....

Melkiades:
> Thanks @Melkiades ! Do you think there is a way we can add a unit test for this? Also, can we add a note in the help file for `as_gt()` that for PDF output, the `\n` line break is stripped due to a bug? I think we should also add a date, so we could perhaps test it again in the future, say after a year or so....

Added a note about it with the date. Let me know if it works ;)

ddsjoberg (approved):
Thank you so much @Melkiades !!!

PR #1925, opened 2024-08-25
Adding `gather_ard()` function

### LINKED ISSUE #1866: Add a `get_ard()` function
This function would extract the ARDs from the gtsummary table and optionally stack them.

ddsjoberg:
@edelarua @Melkiades @ayogasekaram @Nolan-Steed

Can I get your opinion on the name of this function? It's a function that collects all the ARDs used to create a table. It's currently named `get_cards()`, but I think maybe there is a better name?

Here are some alternatives (but open to others as well):
- `get_ard()`: I like this because it has name symmetry for other functions like `cards::bind_ard()`, and every `ard_*()` function.
- `gather_cards()`: A part of me doesn't like the `get_*()` prefix because it reminds me of the `get()` function which gets objects from other enviornments.
- `gather_ard()`: This is just the combination of the two :)

What do you think? I think I am leaning towards `get_ard()`...

* Adding the `gather_ard()` function which collects the ARD data frames used to construct a table. (#1866)

closes #1866

- [ ] Ensure all package dependencies are installed: `renv::install()`
- [ ] PR branch has pulled the most recent updates from master branch: `usethis::pr_merge_main()`

- [ ] Approve Pull Request


edelarua:
A hard decision!! I think any of the 4 would work fine, but I lean slightly more toward the word `gather` rather than `get` since it is more descriptive and for the reason you mentioned. I don't think it will make much of a difference to users since they'll likely be looking up the function anyways :)

Nolan-Steed:
I think I would vote for `gather_ard` for the same reason as it is a little more descriptive and to me it seems like a more unique name so it might be easier for users to remember.

ddsjoberg:
`gather_ard()` it is! Thank you for the input!

R/gather_ard.R
```diff
@@ -0,0 +1,41 @@
+#' Extract ARDs
+#'
+#' Extract the ARDs from a gtsummary table. If needed, results may be combined
+#' with `cards::bind_ard()`.
+#'
+#' @param x (`gtsummary`)\cr a gtsummary table
```
Melkiades, line 6:
```suggestion
#' @param x (`gtsummary`)\cr a gtsummary table.
```

R/gather_ard.R
```diff
@@ -0,0 +1,41 @@
...
+#'
+#' @return list
+#' @export
+#'
+#' @examplesIf gtsummary:::is_pkg_installed('cardx', reference_pkg = 'gtsummary')
```
Melkiades, line 11: shouldnt it be present anyhow?

ddsjoberg: Nearly 100% of the time, yes. But there is a special type of R CMD Check that runs without using any of the Suggested packages, and we're meant to fail gracefully in these situations. So in this case, we just don't run the examples that require any suggested packages.

R/gather_ard.R
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  # the ARD for regression models is an additional calculation, so we don't do it by default
+  if (inherits(x, "tbl_regression")) {
+    check_pkg_installed("cardx", reference_pkg = "gtsummary")
+    return(cardx::ard_regression(x$inputs$x))
+  }
+
+  # grab ARD from standard place
+  if (!is_empty(x[["cards"]])) return(x[["cards"]]) # styler: off
```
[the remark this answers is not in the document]
ddsjoberg: Sometimes I like the single line when I find it easier to read 😉

Melkiades (approved):
Lgtm! Thanks Daniel ;)

ddsjoberg:
> Lgtm! Thanks Daniel ;)
> 
> (`gather_ard` is a good name for me)

Thanks @Melkiades !

Issue #1614, opened 2024-03-19
Add an `add_ard_stat()` function?


ddsjoberg:
This could be very helpful for people building tables from ARDs. This function would convert ARD structures to gtsummary, that is, essentially an API for the reshaping and formatting of the stats in an ARD.

PR #1938, opened 2024-08-28
Adding `add_overall.tbl_ard_summary()`

### LINKED ISSUE #1848: Add `add_overall.tbl_ard_summary()` S3 method
this is so common, it'll be helpful to have

ddsjoberg:
* Adding the `add_overall.tbl_ard_summary()` function. (#1848)

closes #1848

- [x] Ensure all package dependencies are installed: `renv::install()`
- [x] PR branch has pulled the most recent updates from master branch: `usethis::pr_merge_main()`

- [ ] Approve Pull Request


R/add_overall_ard.R
```diff
@@ -0,0 +1,87 @@
...
+#'     .attributes = TRUE,
+#'     .total_n = TRUE
+#'   )
+#'
+#' # add an overall column
+#' tbl |>
+#'   add_overall(cards = ard_overall)
+NULL
+
+#' @rdname add_overall_ard
```
Melkiades, line 57: Why is it on a different page even if it is a S3? I guess to avoid having too many entries? if it is so it is fine for me

R/add_overall_ard.R
```diff
@@ -0,0 +1,87 @@
...
+#'     cards::ard_continuous(variables = age),
+#'     cards::ard_categorical(variables = grade),
+#'     .missing = TRUE,
+#'     .attributes = TRUE,
+#'     .total_n = TRUE
+#'   )
+#'
+#' # add an overall column
+#' tbl |>
+#'   add_overall(cards = ard_overall)
```
Melkiades, line 54: It seems that this function is more adding one column than overall, because you need to specify that column with ard_stack definition. Would it be better to have one add_column function with this structure and one add_overall that takes the relevant info from input ard? like instead of doing tbl |> add_overall more doing ard_stack |> add_overall |> tbl_ard_summary.
Maybe both options?

R/add_overall_ard.R
ddsjoberg: Since 99.9% of gtsummary users don't care about ARDs and the ARD functions have an additional argument not used by the other `add_overall()` methods, I thought it can live in it's own space. Over time, we'll be adding other `tbl_ard_*` `add_overall()` methods.

R/add_overall_ard.R
ddsjoberg: A general function for adding statistics from ARD is on the roadmap. (#1614 )

Also, for easier use, I think we'll also make this easier directly from `tbl_ard_summary()` (#1940)

Melkiades (approved):

PR #2054, opened 2024-11-03
fix: Avoid stray argument

### LINKED ISSUE #2050: Checks fail with duckplyr
The duckplyr package is aimed to be a drop-in replacement for dplyr, with full behavior compatibility. To assert that, I'm running checks with a rigged version of dplyr. This package fails its checks in this scenario.

Details: https://github.com/krlmlr/dplyr/blob/f-revdep-duckplyr/revdep/problems.md .

Learn more about duckplyr: https://duckplyr.tidyverse.org/ .

From the error message, I can't tell immediately what the cause of the failure is. I'd appreciate your help: can you please help digest a reproducible example that shows how duckplyr is behaving differently from dplyr in your use case?

The modified dplyr version can be installed with any of:

```r
pak::pak("krlmlr/dplyr@f-revdep-duckplyr")
# remotes::install_github("krlmlr/dplyr@f-revdep-duckplyr")
# devtools::install_github("krlmlr/dplyr@f-revdep-duckplyr")
```

Thanks a lot for your help! Please let me know if you have any questions.

Tracker: https://github.com/tidyverse/duckplyr/issues/297.

krlmlr:
Closes #2050.


R/brdg_hierarchical.R
```diff
@@ -308,7 +308,6 @@ pier_summary_hierarchical <- function(cards,
     # this ensures the correct order when there are 10+ groups
```
ddsjoberg, line 308: @edelarua can you take a look at this?

- I think we may just need to replace the preceding `|>` with `%>%` to put the input data frame in thte second argument.
- If we don't already have a test for AE tables with 10+ by levels, can we add one?

krlmlr: With both `|>` or `%>%`, the LHS comes in as the first argument, and `cards_no_attr |> ...` as the second. The dot comes in third, and is mapped to `copy`, which currently is never touched in dplyr and hence never errors.

ddsjoberg: Using the dot with `%>%` would change the default location of the argument placement. From the `%>%` help file:

image

krlmlr: Oh, right, I forgot about that. Can we use `y = _` with the base pipe too?

krlmlr: How about storing the RHS in a variable?

ddsjoberg (approved):
Thank you @krlmlr and @edelarua for the update!

PR #2168, opened 2025-02-24
Adding `add_difference_row()`

### LINKED ISSUE #2138: Add `add_comparison_row()` function for multiple comparisons to a single reference group
This function would add a new row chunk. The new rows will have would have a comparison to a single group

Image

https://insightsengineering.github.io/tlg-catalog/stable/tables/efficacy/ttet01.html

ddsjoberg:
* Adding new function `add_difference_row()`. The function is similar to `add_difference()`, except that differences are placed on the rows below the summary statistics. (#2138)

closes #2138

- [ ] Approve Pull Request


ddsjoberg:
@Melkiades can you please review this PR? Thank you!!!

Melkiades (commented):
I did initial testing and everything seems to work great. Just wondering if the reference footnote here is correct:
image

It seems oddly positioned there as the % is missing

R/add_difference.R
```diff
@@ -85,7 +85,7 @@ add_difference.tbl_summary <- function(x,
                                        ...) {
   set_cli_abort_call()
   # check/process inputs -------------------------------------------------------
-  check_dots_empty()
+  check_dots_empty(call = get_cli_abort_call())
```
Melkiades, line 88: isnt it set just above?

Melkiades: is it because there is the generic?

ddsjoberg: > isnt it set just above?

Yep, this is set above, and the `check_dots_empty()` row is a check that throws an error when dots are not empty and we're telling the function which env to reference in the error.

ddsjoberg:
> I did initial testing and everything seems to work great. Just wondering if the reference footnote here is correct: image
> 
> It seems oddly positioned there as the % is missing

That is a good point. I'll update the example to make the footnote more clear

R/add_difference_row.R
```diff
@@ -0,0 +1,435 @@
...
+
+#' Add differences rows between groups
+#'
+#' @description
+#' `r lifecycle::badge('experimental')`\cr
+#' Adds difference to tables created by [`tbl_summary()`] as additional rows.
+#' This function is often useful when there are more than two groups to compare.
+#'
+#' Pairwise differences are calculated relative to the specified
+#' `by` variable's specified reference level.
```
Melkiades, line 26:
```suggestion
#' Pairwise differences are calculated relative to the specified
#' `by` variable's reference level.
```

R/add_difference_row.R
```diff
@@ -0,0 +1,435 @@
...
+#'
+#' - [`add_difference_row.tbl_summary()`]
+#'
+#' @param x (`gtsummary`)\cr
+#'   Object with class 'gtsummary'
+#' @param ... Passed to other methods.
+#' @keywords internal
+#'
+#' @author Daniel D. Sjoberg
+#' @export
```
Melkiades, line 11: if it exported, it should not have the internal keyword, right?

R/add_difference_row.R
```diff
@@ -0,0 +1,435 @@
...
+#' @description
+#' `r lifecycle::badge('experimental')`\cr
+#' Adds difference to tables created by [`tbl_summary()`] as additional rows.
+#' This function is often useful when there are more than two groups to compare.
+#'
+#' Pairwise differences are calculated relative to the specified
+#' `by` variable's specified reference level.
+#'
+#' @inheritParams add_difference.tbl_summary
+#' @param reference (scalar)\cr
```
Melkiades, line 29: maybe better `string`? Can it not be a string?

R/add_difference_row.R
```diff
@@ -0,0 +1,435 @@
...
+#'   Value of the `tbl_summary(by)` variable value that is the reference for
+#'   each of the difference calculations.
+#' @param header (`string`)\cr
+#'   When supplied, a header row will appear above the difference statistics.
+#' @param statistic ([`formula-list-selector`][syntax])\cr
+#'   Specifies summary statistics to display for each variable.  The default is
+#'   `everything() ~ "{estimate}"`. The statistics available to include will
+#'   depend on the method specified in the `test` argument, but are generally
+#'   `"estimate"`, `"std.error"`, `"parameter"`, `"statistic"`,
+#'   `"conf.low"`, `"conf.high"`, `"p.value"`
```
Melkiades, line 39:
```suggestion
#'   `"conf.low"`, `"conf.high"`, `"p.value"`.
```

R/add_p.R
```diff
@@ -509,11 +428,105 @@ calculate_and_add_test_results <- function(x, include, group = NULL, test.args,
...
+
+  # if we captured a warning and the result is an ARD, add the warning to the ARD
+  if (!is.null(lst_captured_results[["warning"]]) &&
+      inherits(lst_captured_results[["result"]], "card")) {
+    lst_captured_results[["result"]][["warning"]] <-
+      lst_captured_results[["result"]][["warning"]] |>
+      map(~c(.x, lst_captured_results[["warning"]]))
+  }
+
+  # if the result is null, replace is with a data frame of of NA
```
Melkiades, line 481:
```suggestion
  # if the result is null, replace it with a data frame of of NA
```

R/add_difference_row.R
```diff
@@ -0,0 +1,435 @@
...
+  if (is_empty(x$inputs$by)) {
+    "Cannot run {.fun add_difference_row} when {.code tbl_summary()} does not include a {.arg by} argument." |>
+      cli::cli_abort(call = get_cli_abort_call())
+  }
+
+  # if `pvalue_fun` not modified, check if we need to use a theme p-value
+  if (missing(pvalue_fun)) {
+    pvalue_fun <-
+      get_theme_element("pkgwide-fn:pvalue_fun") %||%
+      pvalue_fun
```
Melkiades, line 137: Can it remain missing if there is nothing in the theme?

R/add_p.R
```diff
@@ -509,11 +428,105 @@ calculate_and_add_test_results <- function(x, include, group = NULL, test.args,
...
+  if (is.null(lst_captured_results[["result"]])) {
+    lst_captured_results[["result"]] <-
+      rep_named(chr_expected_stats, list(NA)) |>
+      as.data.frame()
+  }
+
+  # check class of returned object. Object must be ARD of class 'card' or a data frame with one row
```
Melkiades, line 488: So the custom function needs to use cards' model constructor right?

R/utils-add_p_tests.R
```diff
@@ -508,7 +508,7 @@ add_p_test_ancova_lme4 <- function(data, variable, by, group, conf.level = 0.95,
               is_empty(adj.vars) ~ list("One-way ANOVA with random intercept"),
               TRUE ~ list("ANCOVA with random intercept")
             ),
-          fmt_fun = list(NULL)
+          fmt_fn = list(as.character)
```
Melkiades, line 511: fmt_fn is default in cards, while fmt_fun is default in gtsummary right? I wanted to check if it was actually aligned but there are many instances (ofc)

data-raw/gtsummary_tests.csv
```diff
@@ -1,53 +1,53 @@
...
-tbl_continuous,TRUE,FALSE,,lme4,lme4,lme4::glmer,"\(...)  gtsummary:::add_p_test_tbl_summary_to_tbl_continuous(..., test_name = ""lme4"")",FALSE,"lme4::glmer(by ~ (1 \UFF5C group), data, family = binomial) %>% anova(lme4::glmer(variable ~ continuous_variable + (1 \UFF5C group), data, family = binomial))",random intercept logistic regression,
-tbl_continuous,TRUE,FALSE,,stats,ancova,,"\(...)  gtsummary:::add_p_test_tbl_summary_to_tbl_continuous(..., test_name = ""ancova"")",FALSE,lm(continuous_variable ~ variable + adj.vars),ANCOVA,
-tbl_continuous,FALSE,FALSE,,stats,ancova_lme4,,"\(...)  gtsummary:::add_p_test_tbl_summary_to_tbl_continuous(..., test_name = ""ancova_lme4"")",FALSE,"lme4::lmer(continuous_variable ~ variable + adj.vars + (1 \UFF5C group), data)",ANCOVA with random intercept,
\ No newline at end of file
+class,add_p,add_difference,diff_statistic,test_package,test_name,test_fun,fun_to_run,accept_dots,pseudo_code,description,details
```
Melkiades, line 1: Here there are no differences in content (it seems to me). I guess github diff does not work correctly on .csv

Could it make sense to make it a .txt (even if read and used as csv) to have the correct diff?

tests/testthat/_snaps/add_p.tbl_summary.md
```diff
@@ -105,7 +105,7 @@
         ...) mtcars))
     Condition
       Error in `add_p()`:
-      ! The test result object for variable "age" is not the expected structure.
+      ! The result from the `test` argument for variable `age` must be an ARD of class <card> or a data frame with one row.
```
Melkiades, line 108: much clearer!!

tests/testthat/test-add_difference.tbl_summary.R
```diff
@@ -614,12 +620,14 @@ test_that("ordering in add_difference.tbl_summary() with paired tests", {
...
+      add_difference(test = ~"paired.t.test", group = id)
+
+    tbl |>
       modify_column_hide(all_stat_cols()) |>
       as.data.frame()
-  )
-  expect_snapshot(
-    mtcars |>
+  })
+  expect_snapshot({
```
Melkiades, line 629: Great way of doing the snapshot testing, i.e. to add the whole generating code block

tests/testthat/test-add_difference_row.tbl_summary.R
```diff
@@ -0,0 +1,140 @@
...
+  tbl_groups <-
+    trial_group |>
+    select(trt, id, stage, marker, age) %>%
+    tbl_summary(
+      by = trt,
+      missing = "no",
+      include = "age"
+    ) |>
+    add_difference_row(
+      test = list(age = "ancova_lme4"),
```
Melkiades, line 115: Ok I see! I think still the error messaging can guide noobs like me towards this a bit more ;)

R/add_difference_row.R
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  cards::process_selectors(x$inputs$data, group = {{ group }}, adj.vars = {{ adj.vars }})
+  check_scalar(group, allow_empty = TRUE)
+
+  cards::process_formula_selectors(
+    scope_table_body(x$table_body, x$inputs$data[include]),
+    test = test,
+    include_env = TRUE
+  )
```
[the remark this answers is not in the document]
Melkiades: I am doing something wrong here! I need a bit of guidance here:
```r
# Example 1 ----------------------------------
trial |>
  tbl_summary(
    by = grade,
    include = c(age),
    missing = "no",
    statistic = all_continuous() ~ "{mean} ({sd})"
  ) |>
  add_stat_label() |>
  add_difference_row(
    reference = "I",
    statistic = everything() ~ c("{estimate}", "{conf.low}, {conf.high}", "{p.value}"), 
    header = "dunno where",
    test = list(age = "paired.t.test"),
    group = "death",
    conf.level = 0.1#,
    # pvalue_fun = label_style_pvalue(digits = 3)#,
    # estimate_fun = everything() ~ label_style_ratio()
  ) #|> 
  add_p()
```
```
Error in `dplyr::mutate()`:
ℹ In argument: `stat_fmt = pmap(...)`.
Caused by error in `add_difference_row()`:
! There was an error applying the formatting function to statistic "p.value" for variable "age".
ℹ Perhaps try formmatting function `as.character()`? See error message below:
✖ Failed to evaluate the left-hand side of formula 4. Caused by error in `abs()`: ! non-numeric argument
  to mathematical function
```
Is the defalt pvalue formatting function not working for the paired.t.test output?

Melkiades (commented):
Overall fantastic PR! I have some doubts on some inputs and other minors, but practically ready to go

R/add_difference_row.R
Melkiades: I guess you need to actively take it out, but having this inside if(missing()) makes the %||% risky maybe?

R/add_difference_row.R
ddsjoberg: I think it's ok to leave as is, since the reference level does need to be specified.

data-raw/gtsummary_tests.csv
ddsjoberg: Usually, it would only show the difference for a CSV file. It looks like the only difference is an extra line break on the last row, which for whatever reason, is appearing as if every line wac changed.

R/utils-add_p_tests.R
ddsjoberg: When I originally named them, I thought they were the same.... 😭 But now wondering if it's too late to change them to match.

R/add_p.R
ddsjoberg: the result can either me an ARD structure or a data frame with one row (which is returned by `broom::tidy()`)

R/add_difference_row.R
ddsjoberg: Generally, yes. But this is just an S3 generic, and I don't need it on the reference page of the website, so I am hiding it.

R/add_difference_row.R
ddsjoberg: The `missing()` function is named poorly. It's really testing if the default value has been changed by the user. What we are saying here is that if the user has _not_ specified the `pvalue_fun` argument, then we use the theme pvalue function, and when that is not set, we land back on the default value for the argument.

tests/testthat/test-add_difference_row.tbl_summary.R
ddsjoberg: What do you have in mind? This code runs without an error, right?

R/add_difference_row.R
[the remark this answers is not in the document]
ddsjoberg: Ahh, thank you for testing this situation. I moved around the order in which the formatting function is applied so we don't see a premature error that is misleading.

R/add_difference_row.R
ddsjoberg: Nice catch! I updated the documentation indicating to pass the character level for factors, and added a check that a scalar factor has not been passed.

ddsjoberg:
Thanks for the thorough review @Melkiades ! Want to take another look before merging?

R/utils-add_p_tests.R
Melkiades: not in this PR ahah

data-raw/gtsummary_tests.csv
Melkiades: ok then nothing to be done! :)

tests/testthat/test-add_difference_row.tbl_summary.R
Melkiades: yes, sorry I was still talking about the previous case, where I added directly the function wrongly

R/add_difference_row.R
Melkiades: Understood! thanks

ddsjoberg:
Thank you so much @Melkiades !! Best reviews!

PR #2217, opened 2025-04-25
Rework `tbl_split` and addition of vertical split (by cols)

Melkiades:
* Added `tbl_split_by_rows()` and `tbl_split_by_columns()` to split tables horizontally (row-wise) and vertically (column-wise).
* Modified `tbl_split()` to handle both column and row split of tables.

#2216 

@ddsjoberg I think this solution is quite clean as it allows to call the splitting only once and do both (vertical and horizontal). I set first column-wise split as I repute it more relevant than row-wise. It can be changed swiftly if you think it differently.

I changed the ambiguous names of horizontal and vertical splits with by_rows and by_columns that is more direct. We had major issues with horizontal and vertical name confusion and I would only have them in the docs.

Let me know what you think! I have not devised tests yet, but the example shows the functionality. 

- [ ] PR branch has pulled the most recent updates from main branch.
- [ ] **All** GitHub Action workflows pass with a :white_check_mark:

- [ ] Approve Pull Request


NAMESPACE
```diff
@@ -77,6 +77,7 @@ S3method(tbl_regression,stanreg)
 S3method(tbl_regression,survreg)
 S3method(tbl_regression,workflow)
 S3method(tbl_split,gtsummary)
+S3method(tbl_split,list)
```
Melkiades, line 80: or tbl_list?

R/tbl_split.R
```diff
@@ -2,22 +2,47 @@
...
+#'
+#' # split by columns
+#' tbl_list_cols <- tbl |>
+#'   tbl_split(key = c(stat_1, stat_2))
+#'
+#' # take the list of splitted rows and split again by columns
+#' tbl_list_rows_cols <- tbl_list_rows |>
+#'   tbl_split(key = c(stat_1, stat_2))
+#'
+#' @name tbl_split
```
Melkiades, line 45: I moved it here bc I always think it is a bit clearer than having to search for it in the doc above. Feel free to move it back there

R/tbl_split.R
```diff
@@ -65,3 +147,62 @@ print.tbl_split <- function(x, ...) {
   check_dots_empty()
   walk(x, print)
 }
```
Melkiades, line 149: how do you want to deal with this? make the viewer do a side by side print? I would do it as an option maybe?

ddsjoberg:
Thanks @Melkiades ! I like the updated names!

Do you think we need row splitting by row number as well as variable  (also, there is the rare gtsumary table that does not have the variable column)?

What is the utility in keeping `tbl_split()`? I am just seeing another function to maintain 😆

Melkiades:
> Thanks @Melkiades ! I like the updated names!
> 
> Do you think we need row splitting by row number as well as variable (also, there is the rare gtsumary table that does not have the variable column)?

I do not think having this option would be too confusing for users and I expect at a certain point they will wonder why it is not present already. It is also true that they can achieve it with sub-setting? If it is not too complex in implementation (I do not think so), I would add it.

> 
> What is the utility in keeping `tbl_split()`? I am just seeing another function to maintain 😆

I like it being the main handler of list splits, so when we already do not know what is going to happen, it devices to list splitting or to row/cols splits. The code is very clean like this (otherwise we would have two separate generics for `tbl_split_by_rows` and `tbl_split_by_cols` so to accept lists too. If you like this second most I can go on and replace fully tbl_split. It will need a deprecation cycle though

Melkiades:
All inputs are welcome! @ayogasekaram @edelarua ^^

R/tbl_split.R
```diff
@@ -28,10 +53,67 @@ tbl_split <- function(x, ...) {
 
 #' @export
 #' @rdname tbl_split
-tbl_split.gtsummary <- function(x, variables, ...) {
+tbl_split.gtsummary <- function(x, keys, variables, groups, ...) {
   set_cli_abort_call()
   check_dots_empty()
 
+  out_list <- NULL
```
Melkiades, line 60: make two separate functions with no generic, only if()

R/tbl_split.R
```diff
@@ -1,36 +1,71 @@
...
+  # list map -------------------------------------------------------------------
+  if (inherits(x, "list")) {
+    out_list <- map(
+      .x = x,
+      .f = tbl_split_by_rows,
+      variables = {{ variables }}
+    )
+    return(
+      out_list |>
+        structure(class = c("tbl_split", "tbl_split_by_rows", "list"))
```
Melkiades, line 63: added some specific for the happened split (by_rows or by_columns)

R/tbl_split.R
```diff
@@ -56,7 +91,81 @@ tbl_split.gtsummary <- function(x, variables, ...) {
...
+    error_msg = "Each element of the {.arg groups} argument's list must be a {.cls character} vector."
+  )
+
+  # check all variables in groups appear in the table
+  columns_do_not_exist <-
+    unlist(groups) |>
+    setdiff(x$table_styling$header$column)
+  if (!is_empty(columns_do_not_exist)) {
+    cli::cli_abort()
+  }
```
Melkiades, line 135: should I test also these?

Melkiades:
I will add footnotes handling later today or tomorrow ;) Let me know if these updates are good @ddsjoberg

R/tbl_split.R
```diff
@@ -1,66 +1,236 @@
-#' Split gtsummary table
+#' Split gtsummary table bz rows and/or columns
```
Melkiades, line 1:
```suggestion
#' Split gtsummary table by rows and/or columns
```

R/tbl_split.R
```diff
@@ -1,66 +1,236 @@
...
+#' @param keys ([`tidy-select`][dplyr::dplyr_tidy_select])\cr
+#'   columns to be repeated in each table split. It defaults to `first_column(x)`
+#'   if missing (usually label column).
+#' @param groups (list of `character` vectors)\cr
+#'   list of column names that appear in `x$table_body`.
+#'   Each group of column names represent a different table in the output list.
+#' @param footnotes (`string`)\cr
+#'   can be either `"first"`, `"each"`, or `"last"`, to locate global footnotes
+#'   only on the first, in each, or in the last table, respectively. It defaults
+#'   to `"each"`. Reference footnotes are always present wherever they appear.
```
Melkiades, line 24: is there something else? reference footnotes (always by what they reference) source and abbreviation footnotes?

R/tbl_split.R
```diff
@@ -1,66 +1,236 @@
...
+      tbl_list[[i]] |>
+        remove_source_note() |>
+        remove_abbreviation()
+    }
+  }
+  if (caption == "first") {
+    # Changing the call_list does not affect the output, should it be trimmed?
+    for (i in seq(2, length(tbl_list))) {
+      tbl_list[[i]]$table_styling$caption <- NULL
+    }
```
Melkiades, line 218: is this enough?

R/tbl_split.R
```diff
@@ -1,66 +1,236 @@
...
+#' # Handle footnotes
+#'
+#' @name tbl_split_by
 NULL
 
 #' @export
-#' @rdname tbl_split
-tbl_split <- function(x, ...) {
-  UseMethod("tbl_split")
+#' @rdname tbl_split_by
```
Melkiades, line 60: It was conflicting with tbl_split. Is it ok?

R/tbl_split.R
ddsjoberg: Hmmm, maybe we should talk this through? I am not sure what we need.
- For footnotes that appear in headers and spanning headers, users may only want to see this once at the end of the table, right?
- Footnotes that appear in the body _must_ appear in the page the footnote appears: otherwise, it won't appear anywhere. BUT, the footnote numbering will be totally out of whack, because each table is numbered independently. Imagine a two page table, with one footnote in the header for both tables, and each table having a footnote in the table, and the user requests all footnotes to appear in the end only. The first table would have the single footnote that appears in the table body and it would be marked 1. The second page would have the header footnote (marked 1) and the table body footnote (marked 2). This sounds like a big mess.
- Source notes (and abbreviations are a type of source note) and much more easily handled because they are not linked to a cell in the table, and we can easily put them anywhere.

For Roche, perhaps this isn't a big deal because we don't have too many footnotes and rely more heavily on source notes, right? Anyway, let's discuss! Aiyah!

R/tbl_split.R
```diff
@@ -1,66 +1,263 @@
...
-#'   tbl_split(variables = c(marker, grade))
+#' # create standard table
+#' tbl <- tbl_summary(trial, by = trt)
+#'
+#' # split by rows ---------------------------
+#' tbl_list_rows <- tbl |>
+#'   tbl_split_by_rows(variables = c(marker, grade))
+#'
+#' # print column names
+#' gtsummary::show_header_names(tbl)
```
ddsjoberg, line 45: No need to namespace gtsummary here

R/tbl_split.R
ddsjoberg: Aiyah, of course. yeah that is good.

R/tbl_split.R
```diff
@@ -1,66 +1,263 @@
...
+      data = scope_table_body(x$table_body),
+      variables = {{ variables }}
+    )
+    # adding last variable
+    variables <- variables |> union(dplyr::last(x$table_body$variable))
+
+    footnotes <- arg_match(footnotes, values = c("each", "first", "last"))
+    # if (missing(footnotes)) {
+    #   footnotes <- get_theme_element("tbl_summary-arg:footnotes", default = translate_string(footnotes)) # styler: off
+    # }
```
ddsjoberg, line 113: Delete this?

R/tbl_split.R
```diff
@@ -1,66 +1,263 @@
...
+    # caption/footnotes handling -----------------------------------------------
+    if (length(tbl_list) > 1) {
+      tbl_list <- .modify_footnotes_caption(tbl_list, footnotes, caption)
+    }
+  }
+
+  # return list of tbls --------------------------------------------------------
+  # assign class (can't assign gtsummary because of print definitions)
+  tbl_list |>
+    structure(class = c("tbl_split", "tbl_split_by_rows", "list"))
```
ddsjoberg, line 146: The most general classes go at the end, so this should be `c("tbl_split_by_rows", "tbl_split", "list")`

R/tbl_split.R
```diff
@@ -1,66 +1,263 @@
...
+
+    # caption/footnotes handling -----------------------------------------------
+    if (length(tbl_list) > 1) {
+      tbl_list <- .modify_footnotes_caption(tbl_list, footnotes, caption)
+    }
+  }
+
+  # return list of tbls --------------------------------------------------------
+  tbl_list |>
+    structure(class = c("tbl_split", "tbl_split_by_columns", "list"))
```
ddsjoberg, line 222: Same comment as above

R/tbl_split.R
Melkiades: yes! I kept it just because I wondered if we want some theme input on this

R/tbl_split.R
Melkiades: yes that sounds like a mess ahah. I think there is no general truth here; we should ask users what they expect/want. I wonder how to make the API more flexible, and if it makes sense to add options too (like something dedicated to the header footnotes)

R/tbl_split.R
Melkiades: I would keep it like this for now

R/tbl_split.R
Melkiades: I think it is not necessary as they are generic checks ;)

R/tbl_split.R
ddsjoberg: yep!

R/tbl_split.R
ddsjoberg: For themes, I generally don't add the theme elements unless I need them.  I went a bit overboard previously, and prefer to maintain as few of them as possible.

Melkiades: understood! Cleaned up

Melkiades:
@ddsjoberg ready for final review!

R/tbl_split.R
```diff
@@ -1,66 +1,257 @@
...
+  set_cli_abort_call()
+
+  # list map -------------------------------------------------------------------
+  if (inherits(x, "list")) {
+    tbl_list <- map(
+      .x = x,
+      .f = tbl_split_by_rows,
+      variables = {{ variables }}
+    )
+  } else {
```
ddsjoberg, line 98: Instead of wrapping all the code in large `if () else ()` chunks, can you update the first chunk to 

```r
if (inherits(x, "tbl_split")) {
  return(...)
}
```

then we don't need the big `else ()` chunk. Same applies to the column split function

R/tbl_split.R
```diff
@@ -1,66 +1,257 @@
...
+#'
+#' @name tbl_split_by
 NULL
 
 #' @export
-#' @rdname tbl_split
-tbl_split <- function(x, ...) {
-  UseMethod("tbl_split")
+#' @rdname tbl_split_by
+tbl_split_by_rows <- function(x, variables, footnotes = "each", caption = "each") {
```
ddsjoberg, line 88: For the `footnotes` and `caption` arguments, can you list all valid choices in a vector, then use `rlang::arg_match()` to validate selection?

R/tbl_split.R
```diff
@@ -1,66 +1,257 @@
...
+  set_cli_abort_call()
+
+  # list map -------------------------------------------------------------------
+  if (inherits(x, "list")) {
+    tbl_list <- map(
```
ddsjoberg, line 93: When we split a table that has been previously split, we are now returning a nested list of tables.  I think the nested list is going to be more difficult to work with downstream compared to a list with depth of 1. Can you flatten the results?

There is a purrr function `list_flatten()` that helps flatten. But this isn't available because we don't depend on purrr. But I wrote a helper in cards, `cards:::.purrr_list_flatten`

R/tbl_split.R
```diff
@@ -1,66 +1,257 @@
...
+#'   definitions `tbl_split_by_rows` and `tbl_split_by_columns`.
+#'
+#' @details
+#' Run [show_header_names()] to print all column names to split by.
 #'
 #' @examples
-#' tbl <-
-#'   tbl_summary(trial) |>
-#'   tbl_split(variables = c(marker, grade))
+#' # create standard table
```
ddsjoberg, line 37: Can you structure this like the other Examples sections, where we number the examples and print the results?

Also, let's make the table smaller, because when these results are printed on the pkgdown site they will be soooo long.

R/tbl_split.R
```diff
@@ -1,66 +1,257 @@
...
+#'
+#' # split by columns ------------------------
+#' tbl_list_cols <- tbl |>
+#'   tbl_split_by_columns(groups = list("stat_1", "stat_2"))
+#'
+#' # take the list of splitted rows and split again by columns
+#' tbl_list_rows_cols <- tbl_list_rows |>
+#'   tbl_split_by_columns(groups = list("stat_1", "stat_2"))
+#'
+#' # Handle footnotes -------------------------
```
ddsjoberg, line 55: I am still waffling on the best way to handle these footnotes. Let's discuss next time we chat.

ddsjoberg (changes requested):
@Melkiades , this is looking great! Let's set up a meeting to discuss the comments!

R/tbl_split.R
Melkiades: It was like this but I changed to process footnotes only once. If you want only one if() I need to duplicate that handling

R/tbl_split.R
Melkiades: I understand! What should we do with the different split object? Does not make sense to leave it to the user to flatten? I can add a comment in the description?

R/tbl_split.R
Melkiades: good point

R/tbl_split.R
Melkiades: done

ddsjoberg (changes requested):
THank you @Melkiades !

I think we just need one more change: Can you add a `row_numbers` argument to `tbl_split_by_rows()`. We need to have the ability to split tables by row numbers as not all tables have a variable column.

- [x] Add the `row_numbers` argument
- [x] Document the `variables` and `row_numbers` arguments together, e.g. `@param variables,row_numbers <description>`
- [x] Add check that a user passed one and only one of `variables` and `row_numbers` argument.

Thanks!

R/tbl_split.R
```diff
@@ -1,36 +1,125 @@
...
+#' @rdname tbl_split_by
+tbl_split_by_rows <- function(x, variables,
+                              row_numbers,
+                              footnotes = c("all", "first", "last"),
+                              caption = c("all", "first")) {
   set_cli_abort_call()
-  check_dots_empty()
+
+  # list map -------------------------------------------------------------------
+  if (inherits(x, "list")) {
```
ddsjoberg, line 107: Shouldn't we be checking for class `tbl_split` here? is there an advantage to keeping it more generic?

R/tbl_split.R
```diff
@@ -40,27 +129,190 @@ tbl_split.gtsummary <- function(x, variables, ...) {
   # adding last variable
   variables <- variables |> union(dplyr::last(x$table_body$variable))
 
+  # footnotes and caption check
+  footnotes <- rlang::arg_match(footnotes)
+  caption <- rlang::arg_match(caption)
```
ddsjoberg, line 134: We don't need to namespace rlang, because we import it.

R/tbl_split.R
```diff
@@ -40,27 +129,190 @@ tbl_split.gtsummary <- function(x, variables, ...) {
...
+    tbl_list <- .modify_footnotes_caption(tbl_list, footnotes, caption)
+  }
+
+  # return list of tbls --------------------------------------------------------
+  tbl_list |>
+    structure(class = c("tbl_split", "list"))
+}
+
+# helper function for handling caption/footnotes
+.modify_footnotes_caption <- function(tbl_list, footnotes, caption) {
```
ddsjoberg, line 281: Let's name this something more specific to splitting..perhaps `.modify_split_footnotes_caption()`, so it's not confused with other similarly named generic functions

R/tbl_split.R
```diff
@@ -40,27 +129,190 @@ tbl_split.gtsummary <- function(x, variables, ...) {
...
+  # footnotes and caption check
+  footnotes <- rlang::arg_match(footnotes)
+  caption <- rlang::arg_match(caption)
+
+  # check that row_numbers is integerish and in bounds
+  if (!is_missing(row_numbers)) {
+    check_integerish(row_numbers)
+
+    # check that row_numbers are in bounds
+    if (any(row_numbers > nrow(x$table_body) | row_numbers < 1)) {
```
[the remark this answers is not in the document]
ddsjoberg: If that function is not in the package, we can update the check function script with `usethis::use_standalone("insightsengineering/standalone", file = "checks")`

R/tbl_split.R
```diff
@@ -40,27 +129,190 @@ tbl_split.gtsummary <- function(x, variables, ...) {
...
+}
+
+#' @export
+#' @rdname tbl_split_by
+first_column <- function(x) {
+  set_cli_abort_call()
+  check_class(x, "gtsummary")
+
+  x$table_styling$header[!x$table_styling$header$hide, ][["column"]][1]
 }
```
ddsjoberg, line 312: I don't think we need to export this function. We have an internal function in gtsummary called `.first_unhidden_column()` with the same functionality, and the user never sees the calls to the function.

tests/testthat/test-tbl_split.R
```diff
@@ -1,9 +1,9 @@
 skip_on_cran()
```
ddsjoberg, line 1: Can you add comments above each of the `expect_*()` function describing the test?

ddsjoberg (changes requested):
Thanks @Melkiades !!! Just a few minor comments. Looking great!

R/tbl_split.R
```diff
@@ -40,27 +129,190 @@ tbl_split.gtsummary <- function(x, variables, ...) {
...
+      )
+    }
+
+    # Check if variables have been provided too
+    if (length(variables) > 1) {
+      cli::cli_abort(
+        "Please select only one between {.arg row_numbers} and {.arg variables} arguments.",
+        call = get_cli_abort_call()
+      )
+    }
```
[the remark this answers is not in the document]
Melkiades: I will make them NULL and use `is_empty`

R/tbl_split.R
Melkiades: I was thinking that some people may want to use their custom lists of different tables

R/tbl_split.R
Melkiades: Right!!

R/tbl_split.R
Melkiades: Very good point. It makes sense! Modified

R/tbl_split.R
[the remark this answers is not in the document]
Melkiades: It is present. Corrected

R/tbl_split.R
Melkiades: done!

PR #2219, opened 2025-04-26
Adding `na` argument to style functions

ddsjoberg:
* The `style_*()` and `label_style_*()` functions have gained the `na` argument allowing users to specify how `NA` values are returned.

- [ ] Approve Pull Request


tests/testthat/test-style_pvalue.R
```diff
@@ -31,6 +31,11 @@ test_that("style_pvalue() works", {
     c(NA, 'p>0.999', 'p=0.999', 'p=0.500', 'p=0.250', 'p=0.200', 'p=0.120', 'p=0.100', 'p=0.060', 'p=0.030', 'p=0.002', 'p<0.001', 'p<0.001', NA)
   )
 
+  expect_equal(
+    style_pvalue(NA, na = "NE"),
+    "NE"
+  )
+
```
ayogasekaram, line 38: Hey @ddsjoberg This function works great when the NA value is explicit in the input. How did you want this behavior to apply if the input value results in an NA value? For example, the test above this one styles 1.5 and -1 to NA - are these NAs to be converted to "NE" as well?

ddsjoberg: Ahhh, that is great point!

So you are saying that inputs that are NA can be recoded to `"NE"`, for example, but if an invalid input is passed and we return an NA, that should probably remain NA, right? I agree! This will have implications for the pvalue, percent, and maybe the ratio function (need to double check on that one)

ayogasekaram: In that case all of the style functions already behave this way. I can add some tests for this specifically in all those files so we can keep track of it :)

ddsjoberg: Yes, that would be great! Thank you!

PR #2292, opened 2025-08-16
No longer calculating `deff` by default in `tbl_svysummary(type='categorical')`

ddsjoberg:
* No longer calculating `deff` by default in `tbl_svysummary(type='categorical')`.

- [ ] Approve Pull Request

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: gtsummary
 Title: Presentation-Ready Data Summary and Analytic Result Tables
-Version: 2.3.0.9002
+Version: 2.3.0.9003
 Authors@R: c(
     person("Daniel D.", "Sjoberg", , "danield.sjoberg@gmail.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0003-0862-2018")),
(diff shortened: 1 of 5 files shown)


ddsjoberg:
> It seems the example in [r-causal/halfmoon#16](https://github.com/r-causal/halfmoon/issues/16) still produces warnings with this change + the CRAN version of cardx. Since that version of cardx hardcodes `deff <- TRUE`, won't it always calculate it anyway?

Thanks @malcolmbarrett ! In `tbl_summary()` for categorical variables, we always calculate `n`, `p`, and `N` even if the user is only include `p` in the table, for example. I did that so we could report the other stats in functions like `inline_text()` (most commonly in my work, I would add `"{n} ({p}%)"` in the table and reported `"{n} / {N} ({p}%)"` in my manuscript.

The `tbl_svysummary()` function was similar in that it would calculate all the stats include `deff`, even if was not reported in the table. I was slightly hopeful that by changing this to only return `deff` when it's requested by the user (the update in this PR). But it looks like we need both changes. 

Thank you for the update in the pkgs!!

PR #2339, opened 2025-10-28
Fix bug in `sort_hierarchical()` when not all variables in `include`

### LINKED ISSUE #2332: Bug Report: sort_hierarchical did not work in partially included and 3+ levels hierarchical table
There is an example:

```
ADAE_subset <- cards::ADAE |>
  dplyr::filter(
    AESOC %in% unique(cards::ADAE$AESOC)[1:5],
    AETERM %in% unique(cards::ADAE$AETERM)[1:5]
  )

# Example 1 - Event Rates --------------------
tbl_hierarchical(
  data = ADAE_subset,
  variables = c(AESOC, AETERM, AESEV),
  include = AESEV,
  by = TRTA,
  denominator = cards::ADSL |> mutate(TRTA = ARM),
  id = USUBJID,
  digits = everything() ~ list(p = 1),
  overall_row = TRUE,
  label = list(..ard_hierarchical_overall.. = "Any Adverse Event")
)
```

which is a 3 levels hierarchical table, but only reporting values in lowest level `AESEV`:

Image

Then, I want the table to be sort by its frequency by piping sort_hierarchical():

```
ADAE_subset <- cards::ADAE |>
  dplyr::filter(
    AESOC %in% unique(cards::ADAE$AESOC)[1:5],
    AETERM %in% unique(cards::ADAE$AETERM)[1:5]
  )

# Example 1 - Event Rates --------------------
tbl_hierarchical(
  data = ADAE_subset,
  variables = c(AESOC, AETERM, AESEV),
  include = AESEV,
  by = TRTA,
  denominator = cards::ADSL |> mutate(TRTA = ARM),
  id = USUBJID,
  digits = everything() ~ list(p = 1),
  overall_row = TRUE,
  label = list(..ard_hierarchical_overall.. = "Any Adverse Event")
) |> 
  sort_hierarchical("descending")
```

Then things go wrong:

Image

I tried to only use 2 variables, things work fine, until I add the third variable... I suspect that this could be potential issue?

edelarua:
* Fixed bug in `sort_hierarchical()` causing variable labels to disappear from table when not all variables are in `include`. (#2332)

Closes #2332 

- [ ] PR branch has pulled the most recent updates from main branch.
- [ ] **All** GitHub Action workflows pass with a :white_check_mark:

- [x] Update `NEWS.md` with the changes from this pull request under the heading "`# gtsummary (development version)`". If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [x] Increment the version number using `usethis::use_version(which = "dev")` 
- [ ] Approve Pull Request
- [x] Merge the PR. Please use "Squash and merge".

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,8 @@
 
 * Corrected handling in `tbl_summary()` of character vectors that contains `"NULL"` values. (#2345)
 
+* Fixed bug in `sort_hierarchical()` causing variable labels to disappear from table when not all variables are in `include`. (#2332)
+
 # gtsummary 2.4.0
 
 ### New Features and Functions


R/sort_hierarchical.R
```diff
@@ -216,36 +216,41 @@ sort_hierarchical.tbl_hierarchical <- function(x, sort = everything() ~ "descend
...
-      set {.code include = everything()} when creating your table."
-    )
+    if ("hierarchical" %in% x$context) { # do not print message for hierarchical count tables
+      for (v in not_incl) {
+        cli::cli_inform(
+          "As {.val {v}} was not specified in {.arg include} the event rates for variable {.val {v}} were estimated by
+          summing the rates for variable {.val {dplyr::last(ard_args$include)}} within each level of {.val {v}}. Due to
+          unique counting of events by {.arg id} these sums may not accurately reflect the true event rates in the data."
+        )
+      }
```
ddsjoberg: I think the message below is enough and a bit shorter.

```suggestion
      cli::cli_inform(
        c("!" = "The rates for {.val {not_incl}} have been {.emph estimated} by
          summing the {.val {dplyr::last(ard_args$include)}} rates.",
          "i" = "Due to unique counting of values within {.val {c(ard_args$id, ard_args$variables)}},
          the summed rates {.emph may not} reflect the true rates.")
      )
```

A couple of things need to be updated

1. The `ard_args` object needs to have the `id` argument value added.
2. The `not_incl` object includes `setdiff(variables, include)`. But should we also intersect this with the variables we are sorting numerically by? (Since users can specify a mix of alphabetical and numeric sorting).


BUT, we don't offer any way for users to get precise sorting. Do you think we need to have a way for more precise sorting? I had two ideas:
1. We could add an argument to `tbl_hierarchical()` to include all rates in the ARD, so we could re-use them for sorting. We would then update this message directing users to that argument and we could avoid estimating rates altogether.
2. We could direct them to include the rates in the primary table and then remove rates they don't want by calling `modify_table_body(\(df) mutate(df, dplyr::across(all_stat_cols(), ~ifelse(variable %in% "AESOC", NA, .))))` after the sort.

R/sort_hierarchical.R
```diff
@@ -216,36 +216,41 @@ sort_hierarchical.tbl_hierarchical <- function(x, sort = everything() ~ "descend
   # add dummy rows for variables not in include so their label rows are sorted correctly
   not_incl <- setdiff(ard_args$variables, ard_args$include)
   if (length(not_incl) > 0) {
-    cli::cli_inform(
-      "Not all hierarchy variables present in the table were included in the {.arg include} argument.
-      These variables ({not_incl}) do not have event rate data available so the total sum of the event rates
-      for this hierarchy section will be used instead. To use true event rates for all sections of the table,
-      set {.code include = everything()} when creating your table."
-    )
+    if ("hierarchical" %in% x$context) { # do not print message for hierarchical count tables
```
ddsjoberg: Instead of `"hierarchical" %in% x$context`, can we use `names(tbl$call_list)`? The context column is a bit weird in some cases, and its future is unknown, while the call list is native to gtsummary.

ddsjoberg (changes requested):
Thank you for this @edelarua ! Just two comments.

R/sort_hierarchical.R
edelarua: Updated.

ddsjoberg:
This is great, thank you so much @edelarua 💪🏼 

Can I get your take on this comment?

> BUT, we don't offer any way for users to get precise sorting. Do you think we need to have a way for more precise sorting? I had two ideas:
> 1. We could add an argument to `tbl_hierarchical()` to include all rates in the ARD, so we could re-use them for sorting. We would then update this message directing users to that argument and we could avoid estimating rates altogether.
> 2. We could direct them to include the rates in the primary table and then remove rates they don't want by calling `modify_table_body(\(df) mutate(df, dplyr::across(all_stat_cols(), ~ifelse(variable %in% "AESOC", NA, .))))` after the sort.

PR #2440, opened 2026-07-31
perf: vectorize tbl_summary() table assembly and remove redundant work

ddsjoberg:

Improve the speed and memory footprint of `tbl_summary()` (and the shared `brdg_summary()`/`pier_*()` bridge, which also backs `tbl_svysummary()`, `tbl_custom_summary()`, `tbl_ard_summary()`, etc.) with **no user-facing change to output** — verified byte-identical via the full snapshot suite.

- Vectorized `pier_summary_categorical()` and `pier_summary_continuous2()`: they now pivot the ARD to wide and interpolate statistics with a single `glue::glue_data()` per variable, mirroring the approach already used in `pier_summary_continuous()`, instead of nested `group_by()`/`group_map()` with per-cell glue. The shared list-column unlisting was extracted into `.unlist_wide_stat_cols()`. The categorical pier preserves the "level-scope stat wins on name collision" semantics of the previous per-level glue.
- Hoisted the `scope_table_body(.list2tb(type, "var_type"), ...)` construction — previously rebuilt 11 times with identical inputs — into two reused locals (`scoped_include`, `scoped_data`), and excluded these internal working objects from the saved `tbl_summary_inputs` so they are not replayed as call arguments (e.g. by `add_overall()`).
- Compute each variable's default dichotomous value once and thread it into `assign_summary_type()` (new backward-compatible optional `dichotomous_values=` argument) and `.assign_default_values()`, instead of computing it twice.
- `Remotes: pharmaverse/cards` builds against the unreleased `cards` performance updates. `main` also carries this entry, so both sides of the CI Performance Benchmark install the same `cards` dev version and the comparison isolates the gtsummary-side changes.

**Benchmark tooling also updated in this PR** (`.github/scripts/benchmark.R`, `.github/workflows/benchmark.yaml`):

- The `tbl_summary` benchmark now exercises all four summary types (age = continuous, marker = continuous2, grade = categorical, response = dichotomous), so every `pier_summary_*()` path is covered.
- Added `main mem` / `pr mem` / `mem Δ` columns (from `bench::mark()`'s `mem_alloc`) to every section of the report, so the memory goal is measured, not just time. Memory allocation is deterministic, so the change is flagged by sign with no confidence interval, and the columns are NA-safe (show `n/a`) when profiling is unavailable.
- The benchmark job now runs in the `rocker/r-ver` container, whose R is built with `--enable-memory-profiling`. The Linux R that `r-lib/actions/setup-r` installs lacks that flag, so `mem_alloc` was `NA` on CI; the container makes the memory columns populate.

**Latest CI Performance Benchmark** (auto-posted as a PR comment and refreshed each run; 5 rounds, 95% CIs, container with memory profiling):

| Benchmark | main | PR | Time (95% CI) | Memory |
|---|---|---|---|---|
| `brdg_summary` (table assembly) | 370.5 ms / 3.05 MB | 291.8 ms / 2.92 MB | ✅ **−21.2%** `[−22.5%, −20%]` | ✅ −4.2% |
| `tbl_summary` pipeline | 1380.9 ms / 13 MB | 1306.9 ms / 13 MB | ✅ −5.3% `[−8.6%, −2.1%]` | ✅ −0.7% |
| `tbl_strata` | 7092.5 ms / 43.9 MB | 6655 ms / 42.3 MB | ✅ −6.2% `[−7.9%, −4.4%]` | ✅ −3.5% |

`brdg_summary()` — the table-assembly stage this PR rewrites — is ~21% faster with a tight CI. The full `tbl_summary` pipeline improves ~5% (CI excludes 0); its wall-clock and memory are diluted by the unchanged `cards` ARD computation, which dominates both. See the auto-posted comment for the full report (style/translation sections included).

N/A

Reviewer notes:
- `NEWS.md` entry added and dev version bumped to `2.5.1.9003`.
- The `Remotes: pharmaverse/cards` entry is intentional and is retained (it is also present on `main`); not a pre-merge removal item.
- The benchmark job now uses `container: rocker/r-ver:latest` to get memory profiling on CI; `setup-r` was dropped since R ships in the image. `latest` keeps main and PR on the same R within a run (fairness preserved); it can be pinned to a fixed version (e.g. `4.6.1`) if reproducibility across time is preferred.
- Correctness guardrail: full `devtools::test()` suite passes with **zero snapshot changes**. The only two failing tests on this branch (`modify_header` deprecation, `tbl_uvregression(method.args)`) were pre-existing on `main`; the `tbl_uvregression` one is fixed in a separate commit here.

### DIFF
--- a/.github/scripts/benchmark.R
+++ b/.github/scripts/benchmark.R
@@ -119,6 +119,7 @@ run_benchmarks <- function(version = "main", n_rounds = 5L) {
       data.frame(
         expression = as.character(all_res$expression),
         median_s = as.numeric(all_res$median),
+        mem_bytes = as.numeric(all_res$mem_alloc),
         round = r,
         version = version,
         pkg_version = pkg_version,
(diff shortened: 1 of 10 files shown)


ddsjoberg:
Hi @kpagacz , this makes the same change you made remove the group map in the a couple of other places. Thanks for your help on the performance improvements!

kpagacz:
Oh, this is fantastic, nicely done! I will be coming back to the performance work soon (might be through the next week), I am working on this in sort of weekly sprints near the end/beginning of a month :) It will still be faster!!

PR #2448, opened 2026-08-01
Fix tbl_strata_nested_stack() misaligned counts for character strata (#2443)

### LINKED ISSUE #2443: Bug Report: Input argument needs to be factored, perhaps adding an assertion to check
We noticed that the output is different when strata by PARAM and PARAMCD for tbl_strata_nested_stack() function. After some research, we found out that when strata for PARAM, the answer is incorrect.

For the following code, I created a simple df and found out the correct answer when strata by PARAM using tbl_summary().

```r
library(gtsummary)
library(dplyr)

# Create sample data with known counts
set.seed(123)
df <- data.frame(
  USUBJID = rep(paste0("SUBJ-", 1:50), each = 4),
  PARAMCD = rep(c("ALB", "ALT", "AST", "BILI"), 50),
  PARAM = rep(c(
    "Albumin (g/L) (SI Result)",
    "ALT (U/L) (SI Result)",
    "AST (U/L) (SI Result)",
    "Bilirubin (umol/L) (SI Result)"
  ), 50),
  ATOXGR = factor(sample(c("0", "1", "2", "3", "4"), 200, replace = TRUE,
    prob = c(0.6, 0.2, 0.1, 0.05, 0.05)
  )),
  TRT01A = factor(rep(c("Drug A", "Drug B"), each = 100))
)

# correct df
df |>
  dplyr::filter(PARAM == "Albumin (g/L) (SI Result)") |>
  tbl_summary(by = TRT01A, include = ATOXGR, type = ATOXGR ~ "categorical")

```

Then, I tried to compare the result of tbl_strata_nested_stack() for factored PARAM and character PARAM, where factored PARAM provides the correct answer.

```
# Character strata: WRONG counts
df |>
  tbl_strata_nested_stack(
    strata = PARAM,
    .tbl_fun = ~ .x |>
      tbl_summary(
        by = TRT01A,
        include = ATOXGR,
        type = ATOXGR ~ "categorical"
      ),
    .header = "Highest Grade"
  )

# Factor strata: CORRECT counts  
df |>
  dplyr::mutate(PARAM = factor(PARAM)) |>
  tbl_strata_nested_stack(
    strata = PARAM,
    .tbl_fun = ~ .x |>
      tbl_summary(
        by = TRT01A,
        include = ATOXGR,
        type = ATOXGR ~ "categorical"
      ),
    .header = "Highest Grade"
  )

```

Hence, the tbl_strata_nested_stack() has a problem when dealing with character strata variable. More specifically, the order when formulating the result table is incorrect. Please check the following two screenshots. We can notice that the count numbers are correct but they are pasted to the wrong section (ALB to ALT, ALT to AST, and AST to ALB).

Image 1Image 2

**In summary, we found out that only factored strata input argument will generate the correct output. We suggest to add an assertion to check for factors.**

ddsjoberg:

* Fixed bug in `tbl_strata_nested_stack()` where summary statistics could be attached to the wrong strata level when the `strata` variable was a character (or other non-factor) vector. (#2443)

closes https://github.com/ddsjoberg/gtsummary/issues/2443

### Details

`tbl_strata_nested_stack()` built two structures from **two independent orderings** and then paired them **positionally**:

- the summary/count tables (`tbls`) — ordered by `cards::nest_for_ard()`
- the header label rows and `tbl_ids` — ordered by `cards::ard_tabulate()` + `dplyr::arrange()`

For **factor** strata both orderings follow the factor level order, and for **numeric** both sort numerically, so they agree. For **character** strata they diverge: `cards::nest_for_ard()` orders with base **C‑locale** collation (e.g. `ALT, AST, Albumin`) while the header pipeline uses **locale‑aware** collation (e.g. `Albumin, ALT, AST`). The mismatch permuted the data relative to the labels, so each strata's statistics were rendered under the wrong strata heading (the cyclic shift reported in #2443).

**Fix:** re‑order `tbls` to match the header ordering by matching on the strata level *values* (`vctrs::vec_match`) instead of relying on positional alignment. The raw strata columns are captured *before* they are replaced by the glued `row_header` strings, so the fix also holds when `row_header` is customized (e.g. `"{strata}, n={n}"`). The factor/numeric code paths are unaffected (the reorder is the identity permutation there).

A unit test was added comparing character‑strata output against the (correct) factor‑strata output for levels whose C‑locale order differs from the locale collation order.

> Note: the underlying divergence is a `cards` inconsistency (`nest_for_ard()` sorting character strata differently than `ard_tabulate()`); filed as pharmaverse/cards#589. This PR makes `tbl_strata_nested_stack()` robust regardless.


ddsjoberg:
Thank you @wzfrank for the detailed report! Can you please take a look at this fix? The sorting mis-match was already addressed in cards's dev version actually. But this code update ensures we no longer rely on an assumed sorting match: we new match the values.

wzfrank:
Hi @ddsjoberg. Thanks for the quick reply and changes. I think this sorting works now!

ddsjoberg:
> Hi @ddsjoberg. Thanks for the quick reply and changes. I think this sorting works now!

Thanks @wzfrank !!

